PR #2, opened 2023-02-13
Add descriptions and licenses to generated packages

gagbo:
# Description

Fix the last few errors to publish the packages.

As everything is published manually for now, there's no
version change here.

Need to add a ticket to make the generators add the missing
keys. (tracked in FP-2995 and FP-2996)

# Checklist

Please make sure all of these are checked before merging. Please leave items
you think are non-applicable in the list, but use strike-through (`~~`) to
indicate they don't apply.

- [x] ~~The changes have been tested to be backwards compatible.~~
- [x] ~~The OpenAPI schema and generated client have been updated.~~
- [x] ~~PR for API is ready and reviewed: (please link)~~
- [x] ~~PR for Studio is ready and reviewed: (please link)~~
- [x] ~~PR for CLI is ready and reviewed: (please link)~~
- [x] ~~PR for Proxy is ready and reviewed: (please link)~~
- [x] ~~The CHANGELOG(s) are updated.~~

When adding new `fiberplane-models` module:

- [x] ~~PR for fp-openapi-rust-gen is ready and reviewed: (please link)~~

When adding new operation types:

- [x] ~~PR for fiberplane-ot is ready and reviewed: (please link)~~

After merging, please merge related PRs ASAP, so others don't get blocked.

PR #3, opened 2023-02-13
chore: move api-client description to correct field

gagbo:
# Description

Small mistake from #2 corrected.

API Client has been published with this change already, to not
block `fiberplane` release, and enable PDK publishing

# Checklist

indicate they don't apply.

- [x] ~~The changes have been tested to be backwards compatible.~~
- [x] ~~PR for Proxy is ready and reviewed: (please link)~~
- [x] ~~The CHANGELOG(s) are updated.~~

When adding new `fiberplane-models` module:

- [x] ~~PR for fp-openapi-rust-gen is ready and reviewed: (please link)~~

### DIFF
--- a/fiberplane-api-client/Cargo.toml
+++ b/fiberplane-api-client/Cargo.toml
@@ -31,13 +31,13 @@ version = "0.3"
 crate-type = ["rlib"]
 edition = "2021"
 name = "fiberplane_api_client"
-description = "Generated API client for Fiberplane API"
 path = "src/lib.rs"
 required-features = []
 
 [package]
 edition = "2021"
 name = "fiberplane-api-client"
+description = "Generated API client for Fiberplane API"
 
 [package.license]
 workspace = true


mellowagain (approved):

arendjr (approved):

arendjr:
Side note: It makes me wonder whether we even should let the OpenAPI generator generate the `Cargo.toml` at all. So much logic in the generator is just for that one file, while it might be easier to just maintain it manually than to update and run the generator :) (@mellowagain)

mellowagain:
> Side note: It makes me wonder whether we even should let the OpenAPI generator generate the Cargo.toml at all. So much logic in the generator is just for that one file, while it might be easier to just maintain it manually than to update and run the generator :) (@mellowagain)

it sort of is an artifact from back when the api client generator was able to generate a fully self containing crate (without referencing fp-models)

but anyways i dont think its _too_ much logic and the code is already there so might as well use it, migrating away from it would also be a decent amount of work as the client generator just fully removes the directory before re-generating so we'd somehow have to save the existing `Cargo.toml` and yada yada

if we fix it up now it should be relatively stable as I dont think well add new dependencies every other week

PR #24, opened 2023-03-07
Add full query state in AutoSuggestionRequest

gagbo:
# Description

This allows the provider to optionally read the context of a completion
request to provide better suggestions

Fixes FP-2731

# Checklist

indicate they don't apply.

- [x] PR for Studio is ready and reviewed: https://github.com/fiberplane/studio/pull/1452

### DIFF
--- a/fiberplane-models/src/providers/auto_suggest.rs
+++ b/fiberplane-models/src/providers/auto_suggest.rs
@@ -14,7 +14,7 @@ use typed_builder::TypedBuilder;
 )]
 #[non_exhaustive]
 pub struct AutoSuggestRequest {
-    /// The query being typed by the user, up to the focus offset.
+    /// The field being typed by the user, up to the focus offset.
     pub query: String,
 
     /// The query type of the provider we're requesting suggestions for.
@@ -22,6 +22,11 @@ pub struct AutoSuggestRequest {
 
     /// The field in the query form we're requesting suggestions for.
     pub field: String,
+
+    /// The current serialized request in its entirety.
+    /// It is not guaranteed to be valid, but each provider can use this to
+    /// provide context-aware suggestions.
+    pub full_query: String,
 }
 
 impl AutoSuggestRequest {
@@ -33,11 +38,13 @@ impl AutoSuggestRequest {
         let mut query = String::new();
         let mut query_type = String::new();
         let mut field = String::new();
+        let mut full_query = String::new();
         for (key, value) in form_urlencoded::parse(&query_data.data) {
             match key.as_ref() {
                 "query" => query = value.to_string(),
                 "query_type" => query_type = value.to_string(),
                 "field" => field = value.to_string(),
+                "full_query" => full_query = value.to_string(),
                 _ => {}
             }
         }
@@ -65,6 +72,7 @@ impl AutoSuggestRequest {
                 query,
                 query_type,
                 field,
+                full_query,
             }),
             false => Err(Error::ValidationError { errors }),
         }
--- a/fiberplane-models/src/providers/auto_suggest.rs
+++ b/fiberplane-models/src/providers/auto_suggest.rs
@@ -23,10 +23,11 @@ pub struct AutoSuggestRequest {
     /// The field in the query form we're requesting suggestions for.
     pub field: String,
 
-    /// The current serialized request in its entirety.
-    /// It is not guaranteed to be valid, but each provider can use this to
-    /// provide context-aware suggestions.
-    pub full_query: String,
+    /// The current query of the cell being filled by the user.
+    /// The string is a URL encoding of the current query (the
+    /// `application/x-www-form-urlencoded` MIME Type is implied.)
+    /// Each provider can use this to provide context-aware suggestions.
+    pub current_query: Option<String>,
 }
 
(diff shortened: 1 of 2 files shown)


fiberplane-models/src/providers/auto_suggest.rs
```diff
@@ -14,14 +14,19 @@ use typed_builder::TypedBuilder;
...
     /// The query type of the provider we're requesting suggestions for.
     pub query_type: String,
 
     /// The field in the query form we're requesting suggestions for.
     pub field: String,
+
+    /// The current serialized request in its entirety.
+    /// It is not guaranteed to be valid, but each provider can use this to
+    /// provide context-aware suggestions.
+    pub full_query: String,
```
arendjr: This field must be optional to avoid a breaking change. (I also don't think it makes sense to make it mandatory semantically.)

I find the name and the doc comment to be somewhat lacking. It's called "full" query, but it may contain partial data. How is that data encoded? Does it contain the `field` and `query` in duplicated form? If so, why?

I think I would go for something like this instead:

```suggestion
    /// If other fields in the provider's query form have already been filled in, this may contain a
    /// form-encoded query string with the data for those fields.
    pub other_field_data: Option<String>,
```

arendjr (changes requested):

fiberplane-models/src/providers/auto_suggest.rs
gagbo: It contains the full content of the cell's data, encoded as a query. There's no special filtering on what data it's supposed to have, so it will likely contain the data in existing fields one additional time. Technically this should be called "query", because it contains the query represented by the cell, but the name query is already taken for the "current field input" data, so I thought "full query" was good.

arendjr (commented):
I still think the naming and comment don't do a great job at contrasting the purpose of `current_query` vs. `query`.

`query` is "The field being typed by the user, up to the focus offset."

`current_query` is "The current query of the cell being filled by the user."

As someone who is not familiar yet with how our provider protoocol works, that immediately raises the question how they differ.  The additional documentation does clarify somewhat, but it still refers to the "current query" which could be easily misunderstood to refer to the `query` field being typed by the user.

This is part of the reason of the reason why I suggested limiting the struct field to only contain data about *other* fields. It doesn't only reduce and deduplicate the data we send, it also makes it much easier to explain the difference between the things we are sending.

Finally, it feels like we are taking a shortcut in the protocol for ease of implementation now that we may come to regret later. Yes, it's easier to send the query data without having to first omit the current field, but that only applies if we can send the query data unaltered. If we look at the definition of the provider cell however, the query data is documented as:

```rust
/// Query data encoded as `"<mime-type>,<data>"`, where the MIME type is
/// either `"application/x-www-form-urlencoded"` or `"multipart/form-data"`.
/// This is used for storing data for the Query Builder.
///
/// Note: The format follows the specification for data URLs, without the
///       `data:` prefix. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs
#[builder(default, setter(strip_option, into))]
#[serde(default, skip_serializing_if = "Option::is_none")]
pub query_data: Option<String>,
```

So we are explicitly keeping open the possibility of using different MIME types. Whenever we decide to actually make use of that (because we implement file upload support in query forms, for instance), we will need to re-encode anyway if we still want to submit a `application/x-www-form-urlencoded`-encoded query string. But at that point we would likely strip files from the fields that we submit for context, and the implied equivalence between this field being "the current query of the cell" and what is really in the cell's query data is broken.

This too I think is a reason for making it explicit that this field *doesn't* necessarily represent the cell's query data.

Sorry for making a rather big deal out of this, but I do really feel that calling it something such as `other_field_data` and omitting the field that we already send will make the functionality easier to explain and more future-proof, with a nice bonus of having to send less data.

fiberplane-models/src/providers/auto_suggest.rs
```diff
@@ -14,14 +14,20 @@ use typed_builder::TypedBuilder;
 )]
 #[non_exhaustive]
 pub struct AutoSuggestRequest {
-    /// The query being typed by the user, up to the focus offset.
+    /// The field being typed by the user, up to the focus offset.
```
arendjr: And to nitpick, I think we better call this "The value of the field", otherwise it is also confusing with the `field` field.

fiberplane-models/src/providers/auto_suggest.rs
```diff
@@ -14,14 +14,20 @@ use typed_builder::TypedBuilder;
...
     pub query_type: String,
 
     /// The field in the query form we're requesting suggestions for.
     pub field: String,
+
+    /// The current query of the cell being filled by the user.
+    /// The string is a URL encoding of the current query (the
+    /// `application/x-www-form-urlencoded` MIME Type is implied.)
+    /// Each provider can use this to provide context-aware suggestions.
+    pub current_query: Option<String>,
```
arendjr: We are still lacking an annotation: `#[serde(default, skip_serializing_if = "Option::is_none")]`

arendjr (approved):
Don't forget to update the field name in the CHANGELOG.

PR #31, opened 2023-03-14
allow setting optional view fields to explicit null

mellowagain:
# Description

views v2 introduces update-able optional fields, which we never had before on any `*_update` route. serde's json parsing does not (and cannot by default) differentiate between the absence of the field and it being explicitly set to `null` with an `Option` marked with `#[serde(default)]`.

this pr introduces a workaround by using a custom deserialize function. source: https://github.com/serde-rs/serde/issues/984#issuecomment-314143738

fixes FP-3079

# Checklist

indicate they don't apply.

- [x] ~~The changes have been tested to be backwards compatible.~~ **this is a breaking change**
- [x] PR for API is ready and reviewed: https://github.com/fiberplane/api/pull/628
- [x] PR for CLI is ready and reviewed: https://github.com/fiberplane/fp/pull/233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ changes if the major version hasn't changed.
 - `fiberplane-models`: All `u32` fields declared within `Pagination` no longer use
   serde's built-in deserialization but a custom visitor. This is a workaround for a
   bug inside axum `Query` <-> serde impl: https://github.com/tokio-rs/axum/discussions/1359 (#27)
+- `fiberplane-models`: `UpdateView` fields `description`, `relative_time`, `sort_by`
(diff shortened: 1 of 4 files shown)


hatchan (approved):

arendjr:
I think the OpenAPI schema no longer matches with the the `UpdateView` struct?

stephlow (approved):

mellowagain:
> I think the OpenAPI schema no longer matches with the the `UpdateView` struct?

i thought at first there was no way to express a optional nullable value but turns out there is: https://stackoverflow.com/a/48114322/11494565

thus, fixed! thanks you

arendjr (approved):

PR #33, opened 2023-03-15
Fetch more history to be able to detect changes

arendjr:
# Description

The [last workflow run](https://github.com/fiberplane/fiberplane/actions/runs/4427147947/jobs/7764340968) for publishing alphas returned:

```
fatal: bad revision ''
🤒 Error: Unexpected exit code (128) from `git diff`
```

I suspect this was because `actions/checkout` doesn't fetch any history by default, so I've bumped the fetch depth to 2. Also made a few small changes to improve the logging in case the issue was something else.

# Checklist

indicate they don't apply.

- ~~[ ] New models module has been added to API generator script~~
- ~~[ ] PR for API is ready and reviewed: (please link)~~
- ~~[ ] PR for Studio is ready and reviewed: (please link)~~
- ~~[ ] PR for CLI is ready and reviewed: (please link)~~
- ~~[ ] PR for FPD is ready and reviewed: (please link)~~
- ~~[ ] The CHANGELOG(s) are updated.~~

- ~~[ ] PR for fiberplane-ot is ready and reviewed: (please link)~~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
 
       - name: setup-git-credentials
         uses: de-vri-es/setup-git-credentials@v2
--- a/fiberplane-ci/src/utils/git.rs
+++ b/fiberplane-ci/src/utils/git.rs
@@ -25,6 +25,7 @@ pub fn get_commits(repo_dir: &str) -> Result<Vec<String>> {
 
     output
         .split(|byte| byte == &b'\n')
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
```
hatchan, line 18: You can also set it to 0 to just grab everything, including branches and tags.

hatchan (approved):

.github/workflows/build.yml
arendjr: Yeah, I don't think I need that much yet, but I'll keep it in mind :)

PR #34, opened 2023-03-16
Decode provider cell query data when converting to a template

gagbo:
# Description

Port of fiberplane/fiberplane-rs#333

Fixes FP-2929

# Checklist

indicate they don't apply.

### DIFF
--- a/fiberplane-templates/src/convert/mod.rs
+++ b/fiberplane-templates/src/convert/mod.rs
@@ -227,45 +227,51 @@ fn print_cell(writer: &mut CodeWriter, cell: &Cell) {
         }
         Cell::Provider(cell) => {
             args.push(("title".to_string(), escape_string(&cell.title)));
-            let cell_type =
-                match cell.intent.as_str() {
-                    "prometheus,timeseries" => {
-                        if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
-                            args.extend(query.into_iter().map(|(key, value)| {
-                                (key.as_ref().to_string(), escape_string(value))
-                            }));
-                        }
-                        "prometheus"
+            let cell_type = match cell.intent.as_str() {
+                "prometheus,timeseries" => {
+                    if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
+                        args.extend(
+                            query
+                                .into_iter()
+                                .map(|(key, value)| (key, escape_string(value))),
+                        );
                     }
-                    "elasticsearch,events" => {
-                        if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
-                            args.extend(query.into_iter().map(|(key, value)| {
-                                (key.as_ref().to_string(), escape_string(value))
-                            }));
-                        }
-                        "elasticsearch"
+                    "prometheus"
+                }
+                "elasticsearch,events" => {
+                    if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
+                        args.extend(
+                            query
+                                .into_iter()
+                                .map(|(key, value)| (key, escape_string(value))),
+                        );
                     }
-                    "loki,events" => {
-                        if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
-                            args.extend(query.into_iter().map(|(key, value)| {
-                                (key.as_ref().to_string(), escape_string(value))
-                            }));
-                        }
-                        "loki"
+                    "elasticsearch"
+                }
+                "loki,events" => {
+                    if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
+                        args.extend(
+                            query
+                                .into_iter()
+                                .map(|(key, value)| (key, escape_string(value))),
+                        );
                     }
-                    _ => {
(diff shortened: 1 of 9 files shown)


fiberplane-templates/src/convert/mod.rs
```diff
@@ -198,26 +206,69 @@ fn print_cell(writer: &mut CodeWriter, cell: &Cell) {
...
+                "loki,events" => {
+                    if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
+                        args.extend(
+                            query
+                                .into_iter()
+                                .map(|(key, value)| (key, escape_string(value))),
+                        );
+                    }
+                    "loki"
+                }
```
arendjr: We should double-check the query data in the cells, otherwise we risk generating invalid templates. As a nice bonus, we'll be able to keep the first value in the `args` tuple a `&str`.

Something like:

```suggestion
                "prometheus,timeseries" | "elasticsearch,events" | "loki,events" => {
                    if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
                        for (key, value) in query.into_iter() {
                            match key => {
                                  "query" => args.push(("content", escape_string(value))),
                                  key => println!("Found unexpected key '{key}' in query data for cell with ID '{}'", cell.id()),
                            }
                        }
                        args.extend(
                            query
                                .into_iter()
                                .map(|(key, value)| (key, escape_string(value))),
                        );
                    }
                    cell.intent.split_once(',').unwrap().0
                }
```

fiberplane-templates/src/convert/mod.rs
```diff
@@ -422,3 +475,31 @@ fn parse_template_function_parameters<'a>(
...
+            .as_ref()
+            .strip_prefix(FORM_ENCODED_QUERY_PREFIX)
+            .map(|encoded_query| {
+                encoded_query
+                    .split('&')
+                    .map(|kv| {
+                        if let Some((key, value)) = kv.split_once('=') {
+                            let value = value.replace('+', " ");
+                            (
+                                percent_decode_str(if key == "query" { "content" } else { key })
```
arendjr: I don't think translating the `key` belongs in this function because it's not part of the decoding logic.

fiberplane-templates/src/convert/mod.rs
```diff
@@ -198,26 +207,63 @@ fn print_cell(writer: &mut CodeWriter, cell: &Cell) {
...
+                "prometheus,timeseries",
+                "elasticsearch,events",
+                "loki,events",
+            ];
+
+            args.push(("title".to_string(), escape_string(&cell.title)));
+            let cell_type = match cell.intent.as_str() {
+                intent if INTENTS_WITH_STDLIB_IMPL.contains(&intent) => {
+                    if let Some(query) = decode_provider_cell_query_data(&cell.query_data) {
+                        args.extend(query.into_iter().filter_map(|(key, value)| {
```
arendjr, line 228: Can we use `push()` here? That way we can keep using `&str` in the `args` tuples, which simplifies a lot of the other code that was changed in this PR because of this.

fiberplane-templates/src/convert/mod.rs
```diff
@@ -198,26 +207,63 @@ fn print_cell(writer: &mut CodeWriter, cell: &Cell) {
...
             ("text", cell.read_only)
         }
         Cell::Provider(cell) => {
-            args.push(("title", escape_string(&cell.title)));
-            args.push(("intent", escape_string(&cell.intent)));
-            if let Some(query_data) = &cell.query_data {
-                args.push(("queryData", escape_string(query_data)));
-            }
-            ("provider", cell.read_only)
+            const INTENTS_WITH_STDLIB_IMPL: [&str; 3] = [
```
arendjr: Stdlib? Do you mean the Fiberplane template helpers? I think I'd call it `INTENTS_WITH_TEMPLATE_HELPERS` then :)

fiberplane-templates/src/convert/mod.rs
```diff
@@ -233,12 +267,14 @@ fn print_cell(writer: &mut CodeWriter, cell: &Cell) {
     }
 
     // Print the cell on one line or multiple depending on how many properties it has
-    let first_param = args.first().map(|(name, _)| *name);
+    let first_param = args.first().map(|(name, _)| name);
```
arendjr: These changes are no longer necessary.

arendjr (approved):

fiberplane-templates/src/convert/mod.rs
gagbo: Oh right I didn't catch that

fiberplane-templates/src/convert/mod.rs
gagbo: extend vs. push doesn't matter too much for the `&str` argument, and leaving `extend` here allows to add a warning if we parsed more than one argument in the query_data too, and to keep a more generic helper function `decode_provider_cell_query_data` (instead of baking the name `query` in the prefix we look for like in the original PR) so I think I'll leave the extend as is for now

PR #39, opened 2023-03-21
Add ArrayField variant in schema.

gagbo:
# Description

Implements RFC-72 about composite fields in provider queries.

This boils down to adding a new variant of `QueryField` named `ArrayField`. Its main trait is that it nests a `QuerySchema` in its value, as `element_schema`, with the following semantics (extracted from the docstring)

## ArrayField::element_schema

The schema of the elements inside a row of the array.

### Accessing row fields

The name of each QueryField inside the element_schema can be used as
an indexing key for a field. That means that if `element_schema` contains
a [TextField](crate::providers::TextField) with the name `parameter_name`,
then you will be able to access the value of that field using
`ArrayField::get(i)::get("parameter_name")` for the i-th element.

### Serialization

For example if an array field has this `element_schema`:
```rust,no_compile
ArrayField {
  name: "table".to_string(),
  element_schema: vec![
    TextField::new().with_name("key"),
    SelectField::new().with_name("operator").with_options(&["<", ">", "<=", ">=", "=="]),
    IntegerField::new().with_name("value"),
  ],
..Default::default()
}
```

Then the URL-encoded serialization for the fields is expected to look like this
(line breaks are only kept for legibility):
```txt
 "table[0][key]=less+than&
 table[2][operator]=%3E&
 table[0][operator]=%3C&
 table[2][key]=greater+than&
 table[2][value]=10&
 table[0][value]=12"
```

Note that we are allowed to skip indices, this serialization is expected to
be read as:
```rust,no_compile
assert_eq!(table, vec![
  Row {
    key: "less than".to_string(),
    operator: "<".to_string(),
    value: 12,
  },
  Row {
    key: "greater than".to_string(),
    operator: ">".to_string(),
    value: 10,
  },
])
```

### Required row fields

Any field that is marked as `required` inside `element_schema` makes it
mandatory to create a valid row to the Array Field.

## Extras

Also derive `Clone, PartialEq` from the QueryField variants in order to manipulate those in dependent code and tests downstream

Fixes FP-2850

# Checklist

indicate they don't apply.

- [ ] The changes have been tested to be backwards compatible.
- [ ] PR for providers is ready and reviewed: https://github.com/fiberplane/providers/pull/31
- [ ] The CHANGELOG(s) are updated.

- [x] PR for fiberplane-ot is ready and reviewed: (please link)~~

### DIFF
--- a/fiberplane-models/src/providers/schema/fields/array_field.rs
+++ b/fiberplane-models/src/providers/schema/fields/array_field.rs
@@ -1,0 +1,137 @@
+use crate::providers::QuerySchema;
+#[cfg(feature = "fp-bindgen")]
+use fp_bindgen::prelude::Serializable;
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+/// Defines an array of composite fields.
+///
+/// This is commonly used for arbitrarily long list of (key, value) pairs,
+/// or lists of (key, operator, value) filters.
+#[derive(Clone, Debug, Default, Deserialize, Serialize, TypedBuilder, PartialEq)]
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::providers")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub struct ArrayField {
+    /// Suggested label to display along the form field.
+    pub label: String,
+
+    /// Name of the field as it will be included in the encoded query or config
+    /// object.
+    pub name: String,
+    /// The minimum number of entries the array must have to be valid.
+    ///
+    /// Leaving the minimum_length to 0 makes the whole field optional.
+    #[builder(default = 0)]
+    pub minimum_length: u32,
+    /// The maximum number of entries the array can have and still be valid.
+    ///
+    /// It is None when there is no maximum number
+    #[builder(default, setter(strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub maximum_length: Option<u32>,
+    /// The schema of the elements inside a row of the array.
+    ///
+    /// ### Accessing row fields
+    ///
+    /// The name of each QueryField inside the element_schema can be used as
+    /// an indexing key for a field. That means that if `element_schema` contains
+    /// a [TextField](crate::providers::TextField) with the name `parameter_name`,
+    /// then you will be able to access the value of that field using
+    /// `ArrayField::get(i)::get("parameter_name")` for the i-th element.
+    ///
+    /// ### Serialization
+    ///
+    /// For example if an array field has this `element_schema`:
+    /// ```rust,no_compile
+    /// ArrayField {
+    ///   name: "table".to_string(),
+    ///   element_schema: vec![
+    ///     TextField::new().with_name("key"),
+    ///     SelectField::new().with_name("operator").with_options(&["<", ">", "<=", ">=", "=="]),
+    ///     IntegerField::new().with_name("value"),
+    ///   ],
+    /// ..Default::default()
+    /// }
+    /// ```
+    ///
+    /// Then the URL-encoded serialization for the fields is expected to look like this
+    /// (line breaks are only kept for legibility):
+    /// ```txt
+    ///  "table[0][key]=less+than&
+    ///  table[2][operator]=%3E&
+    ///  table[0][operator]=%3C&
+    ///  table[2][key]=greater+than&
+    ///  table[2][value]=10&
+    ///  table[0][value]=12"
+    /// ```
+    ///
+    /// Note that we are allowed to skip indices, this serialization is expected to
+    /// be read as:
+    /// ```rust,no_compile
+    /// assert_eq!(table, vec![
+    ///   Row {
+    ///     key: "less than".to_string(),
+    ///     operator: "<".to_string(),
+    ///     value: 12,
+    ///   },
+    ///   Row {
+    ///     key: "greater than".to_string(),
+    ///     operator: ">".to_string(),
+    ///     value: 10,
+    ///   },
+    /// ])
+    /// ```
+    ///
+    /// ### Required row fields
+    ///
+    /// Any field that is marked as `required` inside `element_schema` makes it
+    /// mandatory to create a valid row to the Array Field.
+    pub element_schema: QuerySchema,
+}
+
+impl ArrayField {
+    /// Creates a new array field with all default values.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn with_label(self, label: &str) -> Self {
+        Self {
+            label: label.to_owned(),
+            ..self
+        }
+    }
+
+    pub fn with_element_schema(self, schema: QuerySchema) -> Self {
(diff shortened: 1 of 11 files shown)


fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::providers")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub struct ArrayField {
+    /// Suggested label to display along the form field.
+    pub label: String,
```
arendjr, line 20: ~~Please add `#[builder(setter(into))]` annotations for `String` fields.~~

Scratch that, since we shouldn't need `TypedBuilder` here anyway :)

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Name of the field as it will be included in the encoded query or config
+    /// object.
+    pub name: String,
+    /// The minimum number of entries the array must have to be valid.
```
arendjr, line 26: Please insert a newline between fields when they have comments or annotations, otherwise it looks very crowded.

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// The name of each QueryField inside the element_schema can be used as
+    /// an indexing key for a field. That means that if `element_schema` contains
+    /// a [TextField](crate::providers::TextField) with the name `parameter_name`,
+    /// then you will be able to access the value of that field using
+    /// `ArrayField::get(i)::get("parameter_name")` for the i-th element.
+    ///
+    /// ### Serialization
+    ///
+    /// For example if an array field has this `element_schema`:
+    /// ```rust,no_run
```
arendjr, line 50: Does the `no_run` work for you? I remember trying this before in combination with the `rust` syntax and it would still run them, but maybe my formatting was slightly off...

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// # use fiberplane_models::providers::{ArrayField, TextField, SelectField, IntegerField};
+    /// ArrayField::builder()
+    ///   .name("table".to_string())
+    ///   .label("example".to_string())
```
arendjr: When you have added the `setter(into)` annotations, the explicit `to_string()` calls shouldn't be necessary anymore.

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Then the URL-encoded serialization for the fields is expected to look like this
+    /// (line breaks are only kept for legibility):
+    /// ```txt
+    ///  "table[0][key]=less+than&
+    ///  table[2][operator]=%3E&
+    ///  table[0][operator]=%3C&
+    ///  table[2][key]=greater+than&
+    ///  table[2][value]=10&
+    ///  table[0][value]=12"
+    /// ```
```
arendjr, line 73: Why are the indices randomly ordered? Is it just for illustrative purposes that it is allowed, or will it actually by (semi)random at runtime? If it's the former, maybe that should be clarified with a comment. If it's the latter, I would like to change that behavior.

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+    ///     operator: ">".to_string(),
+    ///     value: 10,
+    ///   },
+    /// ]);
+    /// ```
+    ///
+    /// ### Required row fields
+    ///
+    /// Any field that is marked as `required` inside `element_schema` makes it
+    /// mandatory to create a valid row to the Array Field.
```
arendjr, line 109: Great documentation! :+1:

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+#[cfg(feature = "fp-bindgen")]
+use fp_bindgen::prelude::Serializable;
+use serde::{Deserialize, Serialize};
+use typed_builder::TypedBuilder;
+
+/// Defines an array of composite fields.
+///
+/// This is commonly used for arbitrarily long list of (key, value) pairs,
+/// or lists of (key, operator, value) filters.
+#[derive(Clone, Debug, Default, Deserialize, Serialize, TypedBuilder, PartialEq)]
```
arendjr: Please remove the `TypedBuilder` derive, since this type also has an explicit constructor, so it doesn't need two methods of instantiation. I'm also removing them from all the other `*Field` types in my own PR.

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub element_schema: QuerySchema,
+}
+
+impl ArrayField {
+    /// Creates a new array field with all default values.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn with_label(self, label: &str) -> Self {
```
arendjr:
```suggestion
    pub fn with_label(self, label: impl Into<String>) -> Self {
```

I will also update the other field types to make sure their methods use this consistently as well.

(Same goes for the other methods accepting strings)

fiberplane-models/src/providers/schema/fields/array_field.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    pub fn with_minimum_length(self, minimum_length: u32) -> Self {
+        Self {
+            minimum_length,
+            ..self
+        }
+    }
+
+    pub fn with_maximum_length(self, maximum_length: Option<u32>) -> Self {
```
arendjr:
```suggestion
    pub fn with_maximum_length(self, maximum_length: u32) -> Self {
```

Someone who doesn't want a maximum length should simply not call the method ;)

arendjr (approved):
Nice work!

I think there will be a slight conflict with my own PR, but that's not a major issue. It would be great if you could resolve the comments first though, so I will have less to change :)

fiberplane-models/src/providers/schema/fields/array_field.rs
gagbo: no_run still compiles the test, but it doesn't run it. I had to make changes adding dummy code, just so it compiles, but it doesn't try any `assert` or anything, so it doesn't fail even if it should obviously fail

fiberplane-models/src/providers/schema/fields/array_field.rs
gagbo: It's just showing that any order is allowed. In the end, Studio (or any provider customer) is responsible for the serialization of the query. The example here is just illustrative, as there's no code in `fiberplane-models` that serializes a query whose schema could contain an `ArrayField`

arendjr (approved):
Don't forget to update the CHANGELOG.

PR #41, opened 2023-03-27
Introduce webhooks

mellowagain:
# Description

Based on RFC-90 and RFC-91.

Fixes FP-3129
Fixes FP-3128

# Checklist

- [x] New models module has been added to API generator script.

* API: https://github.com/fiberplane/api/pull/642
* FP: https://github.com/fiberplane/fp/pull/239


fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,79 @@
...
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::webhooks")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub struct Webhook {
+    pub id: Base64Uuid,
+    pub workspace_id: Base64Uuid,
+    pub endpoint: String,
```
arendjr, line 81: These (types such as `String` and `Base64Uuid`, and `Timestamp` below) should use `#[builder(setter(into))]`.

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub events: Vec<WebhookCategories>,
+    #[builder(default, setter(strip_option, into))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub shared_secret: Option<String>,
+    pub enabled: bool,
```
arendjr, line 87: This can use `#[builder(default)]`.

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,110 @@
...
+pub struct UpdateWebhook {
+    #[builder(default, setter(strip_option, into))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub endpoint: Option<String>,
+    #[builder(default, setter(strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub events: Option<Vec<WebhookCategories>>,
+    #[builder(default, setter(strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub regenerate_shared_secret: Option<bool>,
```
arendjr: It might make sense to omit the `Option` altogether here, because this is not a field being set on the webhook, but rather an instruction. It's a bit of a special case, I admit, but it might make the model slightly easier to use also.

```suggestion
    #[builder(default)]
    #[serde(default)]
    pub regenerate_shared_secret: bool,
```

gagbo (approved):

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,173 @@
...
+
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::webhooks")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "lowercase")]
```
arendjr: We use `#[serde(rename_all = "snake_case")]` for enums. Could you add this to the `CONTRIBUTING.md` also?

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+// required to be a specialized `impl` because only 0 and 1 are covered cases
+impl Into<WebhookCategory> for i16 {
+    fn into(self) -> WebhookCategory {
+        match self {
+            0 => WebhookCategory::Ping,
+            1 => WebhookCategory::FrontMatter,
+            value => panic!("unknown value {value}, expected 0 (ping) or 1 (frontmatter)"),
```
arendjr: I think we should use `TryFrom` instead of panicking.

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[builder(default, setter(into))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub response_headers: Option<String>,
+    #[builder(default, setter(into))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub response_body: Option<String>,
+
+    #[builder(setter(into))]
+    pub sent_request_at: Timestamp,
+    #[builder(default, setter(into))]
```
arendjr: A lot of `Option`s here that should use `#[builder(strip_option)]`.

schemas/openapi_v1.yml
```diff
@@ -4234,3 +4370,141 @@ paths:
...
+                  $ref: "#/components/schemas/webhook"
+    post:
+      operationId: webhook_create
+      summary: Create a new webhook
+      description: | 
+        Create a new webhook. Upon execution of this route, a test event ("ping") will be sent to the endpoint.
+        If sending the ping event fails, the webhook will still be created, but it will be disabled.
+        Please check the response of this endpoint to see whenever `enabled` = `false`,
+        and if that is the case, check the latest delivery to see why. Once the endpoint has been fixed, you
+        can re-deliver the ping payload and if that works, manually re-enable the webhook again.
```
arendjr, line 4402: Nice docs :+1:

arendjr (changes requested):

fiberplane-models/src/webhooks.rs
mellowagain: the categories for webhooks were spec'ed in the rfc to be completely lowercase, which is why I added this attribute. we can change it to `snake_case` tho, it'll only impact `frontmatter` which will become `front_matter`

arendjr (approved):
:partying_face:

fiberplane-models/src/webhooks.rs
mellowagain: i actually remembered why after having done the change lol

turns out that i need the setters to accept a `Option` as these fields can be null in the database on the api side as well

mellowagain: i added a comment linking to this thread so future readers don't get confused

arendjr: Right. I see why you'd take that route, though personally I think we should just take the pain in the API so the models stay consistent. After all, it's not as if the API cannot use the type anymore, it just becomes less convenient *in that specific instance*. But by making exceptions based on usage in the API, we're effectively creating an inconsistent public API based on which usages are convenient in one of our internal projects.

We have a bunch of other places too that look similar to what we could do here:

```rs
let mut delivery = WebhookDelivery::builder()
    .id(delivery_id)
    .webhook_id(webhook_id)
    // more of these
    .build();
delivery.inconvenient_option = option_from_db;
delivery
```

mellowagain: i mean fair, yea i guess it makes more sense to have a consistent and clean interface publicly than to have a bit of pain in the api privately, i'll revert the commit

arendjr: Thanks!

schemas/webhooks.schema.json
```diff
@@ -0,0 +1,118 @@
+{
```
hatchan, line 1: Where is this used?

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,188 @@
+use crate::timestamps::Timestamp;
+use base64uuid::Base64Uuid;
+#[cfg(feature = "fp-bindgen")]
+use fp_bindgen::prelude::Serializable;
+use serde::{Deserialize, Serialize};
+use strum::IntoEnumIterator;
```
hatchan: Little clippy that can be resolved

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::webhooks")
+)]
+#[error("unknown value {0}, expected 0 (ping) or 1 (front_matter)")]
+pub struct InvalidWebhookIdError(i16);
```
hatchan: Shouldn't this be `InvalidWebhookCategoryError`?

fiberplane-models/src/webhooks.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub endpoint: Option<String>,
+    #[builder(default, setter(strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub events: Option<Vec<WebhookCategory>>,
+    #[builder(default)]
+    #[serde(default)]
+    pub regenerate_shared_secret: bool,
+    #[builder(default, setter(strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub enabled: Option<bool>,
```
hatchan: Maybe we can add some newlines in between the properties to make it a bit more readable.

```suggestion
    #[builder(default, setter(strip_option, into))]
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub endpoint: Option<String>,

    #[builder(default, setter(strip_option))]
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub events: Option<Vec<WebhookCategory>>,

    #[builder(default)]
    #[serde(default)]
    pub regenerate_shared_secret: bool,

    #[builder(default, setter(strip_option))]
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub enabled: Option<bool>,
```

Also applies to the other structs

hatchan (changes requested):

schemas/webhooks.schema.json
mellowagain: nowhere, it is simple here so others can use it

fiberplane-models/src/webhooks.rs
mellowagain: the import is in fact used by the expansion of the `EnumIter` proc macro

hatchan (approved):

PR #48, opened 2023-04-24
Fiberplane Charts

arendjr:
# Description

Adds a separate `fiberplane-charts` package that contains all the charting functionality used in Studio, so that we can start using it inside the VS Code extension for Autometrics as well. This PR is rather special because it adds the first non-Rust dependency to the repository. Because of this, it contains quite some overhead/setup to get Yarn, Rome and TypeScript going in this repository. The actual code matches the chart component from Studio, though it also contains modifications:

* All references to Redux have been removed to keep the component as self-contained as possible.
* There's a new `ChartTheme` type with theming options. Most of these are a subset from the `Theme` in Studio, but I've also added a few chart-specific ones: notably, `fontAxis*`, `fontControls*`, `fontLegend*` and `fontResults*`.
* A handful of utils have been copied from Studio. It's a bit of a shame we have multiple copies of those now, but they're relatively trivial and I think it would be more of a hassle to have to maintain separate packages for them.
* Same goes for the `BaseComponents` in this repo, which are effectively stripped copies of a few of the `UI` components in Studio.
* There's no Storybook or similar setup yet. For now I'm happy it still looks the same in Studio :)

This PR is a prerequisite to: https://github.com/autometrics-dev/vscode-autometrics/issues/12

# Checklist

indicate they don't apply.

- ~~[ ] New models module has been added to API generator script.~~
- ~~[ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~~

* Studio: https://github.com/fiberplane/studio/pull/1500


mellowagain:
why did we decide to put it here into fiberplane instead of a separate repo? i thought this repo only houses our common rust dependencies

arendjr:
Mainly because it has a dependency on the provider types, which are also in this repo. So if we don't put it here, we need to manage even more cross-repo dependencies, which would likely imply forcing us to publish packages, which would be a pretty big drain on our DX again. We could avoid requiring ourselves to publish packages only if we are okay with adding `package.json` files to this repository anyway, so once we take that route, why not accept that we do indeed host TS code beyond just the generated TS runtime here...

mies:
Random question/request for this charting library; could we have a "copy image to clipboard" functionality. This would enable ppl to paste this for instance into slack and other tools?

arendjr:
@mies Added a discussion for it ;) https://github.com/fiberplane/fiberplane/discussions/49

mellowagain:
> Mainly because it has a dependency on the provider types, which are also in this repo. So if we don't put it here, we need to manage even more cross-repo dependencies, which would likely imply forcing us to publish packages, which would be a pretty big drain on our DX again. We could avoid requiring ourselves to publish packages only if we are okay with adding `package.json` files to this repository anyway, so once we take that route, why not accept that we do indeed host TS code beyond just the generated TS runtime here...

valid, yea i agree then we should add it here

mellowagain (approved):
seems fine from repo setup perspective, haven't taken a look at the `.ts[x]` files as most already passed review when they were put into studio, ill leave the new ones for @fiberplane/studio-devs to review :)

flenter (approved):
LGTM, but two questions:

* why is the `dist` folder included in the repo?
* why not indent with 2 spaces like our internal typescript code?

arendjr:
The `dist` folder is required to allow Studio and the VS Code extension to install directly from the repository, instead of requiring us to publish a package first. I don't mind publishing a package *also*, but it would be very cumbersome if publication were a requirement for every change to be usable downstream.

I had to choose between consistency within this repo, versus with TS code elsewhere. The Rome extension wants to format JSON files too, and the repository already had JSON files that used 4 spaces, so I went with that. It can also be configured with an exception for JSON files, so ultimately it's a coin toss... I'd be happy to change it :)

emschwartz (approved):

PR #56, opened 2023-05-15
Make charts more customizable

flenter:
# Description

For the vscode (autometrics) extension charts needed to be made more customizable. Charts needed to:
* no always column lines. in the grid
* grid lines needed to be dotted
* footer shouldn't be visible
* chart controls aren't needed


oscarvz (approved):
👍 nice!

PR #57, opened 2023-05-19
Chart line colors

flenter:
# Description

VS code has its own set of. colors, this change to the charts component means that the default theme colors can be overruled (something that the vscode extension for autometrics will take advantage of).


brettimus (approved):

brettimus:
worked with my test repo that pulls in charts 👍

PR #58, opened 2023-05-24
add models for sign in with github

mellowagain:
# Description

Fixes FP-3228

# Checklist

- [x] ~~New models module has been added to api generator xtask~~

* API: https://github.com/fiberplane/api/pull/658
* Infrastructure: https://github.com/fiberplane/infrastructure/pull/324

### DIFF
--- a/fiberplane-models/src/users.rs
+++ b/fiberplane-models/src/users.rs
@@ -62,3 +62,16 @@ pub struct OidConnection {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub sub: Option<String>,
 }
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize, TypedBuilder)]
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::users")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub struct OidLinkupLocation {
+    #[builder(setter(into))]
+    pub location: String,
+}
--- a/fiberplane-models/src/users.rs
+++ b/fiberplane-models/src/users.rs
@@ -34,17 +34,17 @@ pub struct Profile {
     pub roles: HashMap<Base64Uuid, AuthRole>,
 }
 
-#[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize, TypedBuilder)]
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize)]
 #[cfg_attr(
     feature = "fp-bindgen",
     derive(Serializable),
     fp(rust_module = "fiberplane_models::users")
 )]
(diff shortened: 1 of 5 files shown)


fiberplane-models/src/users.rs
```diff
@@ -33,3 +33,32 @@ pub struct Profile {
...
+    fp(rust_module = "fiberplane_models::users")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub struct OidConnection {
+    pub linked: bool,
+    /// The unique ID on the providers' site for the user, if linked
+    #[builder(default, setter(into, strip_option))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sub: Option<String>,
```
gagbo: Can we also add the `iss` in the struct, so that later when in codebase we have random `OidConnection`s we have a field that also traces back to the issuer to know where the claim comes from? Seems like right now we are relying on always accessing this `OidConnection` through `OidConnections` to remember where the `sub` is registered

gagbo (commented):
It seems fine! I just have a small question about the possibility to add the issuer of OpenID connections on top of the identity

fiberplane-models/src/users.rs
mellowagain: sure

gagbo (approved):

arendjr (approved):

fiberplane-models/src/users.rs
```diff
@@ -33,3 +33,46 @@ pub struct Profile {
...
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::users")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "snake_case")]
+pub enum OidProvider {
+    Google,
+    Github,
```
hatchan:
```suggestion
    GitHub,
```

schemas/openapi_v1.yml
```diff
@@ -2260,6 +2260,30 @@ components:
...
+        - linked
+      properties:
+        provider:
+          type: string
+          enum:
+            - google
+            - github
+        linked:
+          type: boolean
+        sub:
```
hatchan: Isn't this something that is supposed to be kept private? Is there a reason why Studio would need this?

mellowagain: nope its literally just the user id on the providers' side lol

hatchan: > Is there a reason why Studio would need this?

If it isn't required I don't think we should return it

fiberplane-models/src/users.rs
mellowagain: i have implemented the change, but I needed to add a `#[serde(rename = "github")]`, as else the enum member would be serialized as `git_hub`, which is a spelling i really haven't seen anywhere, at all

hatchan (approved):

PR #61, opened 2023-06-12
Bump internal dependencies when bumping workspace version

arendjr:
# Description

The `Cargo.toml` in the root specifies version numbers for internal dependencies, which also need to be updated when changing the workspace version number.

Also contains some other small CI fixes.

### DIFF
--- a/xtask/src/commands/versions.rs
+++ b/xtask/src/commands/versions.rs
@@ -47,6 +47,16 @@ fn handle_set_version(args: &SetVersionArgs) -> TaskResult {
 
         println!("{SUCCESS}Version updated.");
     } else {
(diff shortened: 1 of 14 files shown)


xtask/src/commands/versions.rs
```diff
@@ -47,6 +47,16 @@ fn handle_set_version(args: &SetVersionArgs) -> TaskResult {
 
         println!("{SUCCESS}Version updated.");
     } else {
+        for crate_name in crates_using_workspace_version {
+            set_version(
+                format!("Cargo.toml"),
```
mellowagain: clippy

mellowagain (approved):

PR #64, opened 2023-06-13
Fix legend / series toggling 

flenter:
# Description

Right now the legend items on hover show a pointer cursor and change background hinting that they are interactive however nothing is wired up to really respond to it. This PR updates the types so that the MetricGraph accepts the `onToggleTimeseriesVisibility` allowing something to happen when a user clicks on it. 

Also: when this property isn't set, the legend items no longer will change color on hover, nor will the pointer cursor be displayed.


arendjr (approved):
Looks good!

I do notice the `test_typescript` workflow doesn't run. Would you mind having a look why that is? I thought I had enabled it...

mellowagain (approved):
workflow looks good

PR #67, opened 2023-06-21
Add SparkChart

flenter:
# Description

For the explorer view for autometrics we want to have small sparkcharts. 
image

This PR contains:
* a new SparkChart component
* a refactor for the MetricsChart code: code that's shared between the MetricsChart and the SparkChart now lives in a `CoreChart` folder
*  the ChartSizeContainerProvider is no longer fixed in the height and things like margins (however there's no change in the behaviour for MetricCharts)
* fixes an issue when you had multiple instances/sizes of charts on a page. Before: multiple charts would use the same ids, however the clippath for the charts differ depending on the width/height which typically was never an issue but with the introduction of the SparkCharts this has changed.

# Checklist

   Tested with:
  - [x]  autometrics vscode plugin
  - [x] studio
  - [x] autometrics-explorer
- [ ] The OpenAPI schema and generated client have been updated.
- [ ] New models module has been added to api generator xtask
- [ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).


fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -15,6 +15,7 @@ type Props = {
     yScale: ValueScale;
     xScaleFormatter?: TickFormatter<Date | NumberValue>;
     gridColumnsShown?: boolean;
+    gridRowsShown?: boolean;
```
arendjr, line 18: I see the convention was already there, so no need to change it now, but I just wanted to say it feels rather backwards that properties are named in the past tense when they are effectively instructions to the component for something that still needs to be done :sweat_smile:

arendjr (approved):
Nice!

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
flenter: Right,I think we've had discussions about this in the past. The idea for this style of naming is that it is in line with "checked" and "disabled" plus when glancing at the prop name (without seeing the type) you don't get the impression that it is a function.

PR #71, opened 2023-07-05
`fiberplane-charts`: Split charts lib

arendjr:
# Description

This extracts the "ACG" (abstract chart generator) to become a standalone component inside `fiberplane-charts`. Both the `MetricsChart` and the `SparkChart` call the ACG, while the internal `CoreChart` has been refactored to render abstract charts instead.

This PR also reduces our dependency on VisX slightly, although I only used direct SVG components where doing so was low hanging fruit.

TODO:

- [x] Make sure dragging along the X axis works
- [x] Fix positioning of bar charts
- [x] Fix percentage stacked charts
- [x] Fix tooltips

Implements [RFC 104](https://studio.fiberplane.com/workspaces/fiberplane/notebooks/RFC-104-Splitting-the-charts-library-5H6VTTRoTxmUpSxZERmNVg).

# Checklist

~~- [ ] The changes have been tested to be backwards compatible.~~

- [ ] Monofiber: https://github.com/fiberplane/monofiber/pull/63


fiberplane-charts/src/CoreChart/ChartShape/AreaShape.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <Threshold<AreaPoint<P>>
+        id={id}
+        data={area.points}
+        x={getX}
+        y0={getY0}
+        y1={getY1}
+        clipAboveTo={0}
+        clipBelowTo={getY1}
+        aboveAreaProps={{ fill: fillColor }}
+        // Keep this one around to spot any incorrect threshold computations.
```
brettimus, line 44: Should we remove this in production builds?

fiberplane-charts/src/CoreChart/ChartSizeContainerProvider.tsx
```diff
@@ -31,11 +31,10 @@ export function ChartSizeContainerProvider({
   const intersectionRef = useRef<HTMLDivElement>(null);
   const ref = mergeRefs([measureRef, intersectionRef]);
 
-  const intersections = useIntersectionObserver(intersectionRef, {
-    root: null,
-    rootMargin: "0px",
-    threshold: 0,
-  });
+  const intersections = useIntersectionObserver(
```
brettimus, line 34: aside: i appreciate all the little cleanup stuff you did

fiberplane-charts/src/CoreChart/CoreChart.tsx
```diff
@@ -1,156 +1,157 @@
...
-    </TooltipContext.Provider>
+    // rome-ignore lint/a11y/noSvgWithoutTitle: title would interfere with tooltip
+    <svg
+      width={width}
+      height={height}
+      onMouseDown={onMouseDown}
+      onMouseMove={onMouseMove}
+      onMouseUp={onMouseUp}
+      onMouseLeave={onMouseLeave}
+      style={{ cursor, marginTop: 2 }}
```
brettimus, line 94: do you remember why `marginTop: 2` is here? if so, could you put a small comment for future readers?

fiberplane-charts/src/CoreChart/CoreChart.tsx
```diff
@@ -1,156 +1,157 @@
...
-function getCursorFromState(
-  interactiveControlsState: InteractiveControlsState,
-  shiftKey: boolean,
-): string {
-  switch (interactiveControlsState.type) {
+function getCursorFromState(state: InteractiveControlsState): string {
+  switch (state.mouseInteraction.type) {
     case "none":
-      return shiftKey ? "grab" : "default";
+      if (state.dragKeyPressed) {
```
brettimus, line 145: nice semantic abstraction of the drag and zoom keypresses

fiberplane-charts/src/CoreChart/GridWithAxes/BottomAxis.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <line
+        x1={0}
+        y1={0}
+        x2={xMax}
+        y2={0}
+        stroke={strokeColor}
+        strokeDasharray={strokeDasharray}
+      />
+
+      {getTicks(xScale, numTicks).map((value, index) => (
```
brettimus, line 49: something that the visx library did was add some extra space between the labels and axis. they had this as a default value of `8` on their AxisBottom component (see: [here](https://github.com/airbnb/visx/blob/94d561aa3f75e44b8d35918ea4da0cacd9177d98/packages/visx-axis/src/axis/AxisBottom.tsx#L17), which appears to be used to calculate a transform on the <text> elements [here](https://github.com/airbnb/visx/blob/master/packages/visx-axis/src/utils/getLabelTransform.ts#L14))

you can see the effect in this screenshot (refactored fiberplane-chars):

image

compared to this (old fiberplane-charts):

image

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -130,7 +97,7 @@ function useCustomSpring(value: number) {
   const [current, setCurrent] = useState(value);
 
   useLayoutEffect(() => {
-    return motionValue.onChange((value) => setCurrent(value));
+    return motionValue.on("change", (value) => setCurrent(value));
```
brettimus, line 100: nice, i think this will resolve a warning we were getting in the console

fiberplane-charts/src/CoreChart/GridWithAxes/LeftAxis.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const {
+    colorBase500,
+    fontAxisFontSize,
+    fontAxisFontFamily,
+    fontAxisFontStyle,
+    fontAxisFontWeight,
+    fontAxisLetterSpacing,
+  } = useTheme();
+
+  const tickLabelProps = {
```
brettimus, line 31: same comment as with BottomAxis: it could be nice to add a `labelOffset` of some sort, to put some space between the labels and axis. could be simple for now, a fixed value

fiberplane-charts/src/CoreChart/hooks/useMouseControls.ts
```diff
@@ -0,0 +1,179 @@
...
+      resetMouseInteraction();
+    }
+  }
+
+  function onWheel(event: WheelEvent) {
+    if (mouseInteraction.type !== "none" || !zoomKeyPressed) {
+      return;
+    }
+
+    const coords = getCoordinatesForEvent(event, dimensions);
```
brettimus, line 119: this is throwing errors for me on wheel events. getCoordinatesForEvent assumes that `event.target` is the svg element, and at least for me (latest chrome), it's usually the `window` object. so, when it tries to get the bounding client rect to calculate coordinates, we throw an error. 

the effect is that "scroll to zoom in or zoom out" doesn't work anymore. not a blocker imo, but worth following up on

brettimus: the previous implementation relied on a ref to the chart content:

```ts
 const graphContent = graphContentRef.current;
    if (!graphContent) {
      return;
    }

    const rect = graphContent.getClientRects()[0];
    const x = event.pageX - (rect?.left ?? 0);
    if (x < 0 || x > xMax) {
      return;
    }
```

brettimus (changes requested):
did a quick first pass and found some non-blockers. will continue review tomorrow!

fiberplane-charts/src/CoreChart/ChartShape/AreaShape.tsx
arendjr: We kept it around thus far without trouble, I don't think it would make much difference...

It did actually help me debug an issue I had introduced though :)

fiberplane-charts/src/CoreChart/CoreChart.tsx
arendjr: I have no idea... it was there, but also only for interactive charts and not read-only ones... I could not find a good reason for that either, so now it's always there. I assume it's just part of the design somehow...

fiberplane-charts/src/CoreChart/GridWithAxes/BottomAxis.tsx
arendjr: Good catch, thanks!

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
arendjr: Yeah, it does :)

fiberplane-charts/src/CoreChart/hooks/useMouseControls.ts
arendjr: Oh, great catch! I had fixed the scroll behavior, then moved the element to which the handler was attached to fix an issue where the default behavior could not be stopped, and forgot to check it afterwards.

arendjr:
I think I've addressed all the feedback thus far!

fiberplane-charts/src/Mondrian/generator/utils.ts
```diff
@@ -0,0 +1,325 @@
...
+): Axis {
+  const minMax = getBucketsMinMax(buckets, getMinMax);
+  if (!minMax) {
+    return getYAxisForConstantValue(0);
+  }
+
+  const [minValue, maxValue] = minMax;
+
+  if (minValue === maxValue) {
+    getYAxisForConstantValue(minValue);
```
brettimus: missing return?

brettimus (approved):
fantastic

fiberplane-charts/src/Mondrian/generator/utils.ts
brettimus: i'm 99.99% sure this meant to have a return statement, so i just added it ✨

arendjr: Perfect!

PR #75, opened 2023-07-19
`fiberplane-charts`: Align ticks to buckets

arendjr:
# Description

This aligns ticks along the X axis to the buckets returned by Prometheus/the Provider. This makes for a much smoother experience with the tooltips especially.

Also reduces our dependency on VisX some more, since we don't use their `getTicks()` or grid components anymore.

# Checklist

* Monofiber: https://github.com/fiberplane/monofiber/pull/63


brettimus:
small regression when "dragging" across time. in previous implementation, the grid lines (ticks) would move with the chart (see video)


https://github.com/fiberplane/fiberplane/assets/3201327/70b9fc52-9594-4bac-8fef-cf8b00656a7d

fiberplane-charts/src/Mondrian/generator/utils/attachSuggestionsToXAxis.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Adds suggestions to the axis based on the position of the first bucket and
+ * the interval between buckets.
+ *
+ * @note This function mutates its input axis.
+ */
+export function attachSuggestionsToXAxis(
+  xAxis: Axis,
+  buckets: Buckets<unknown>,
+  interval: number,
```
brettimus, line 14: don't know if this would be overly defensive but should there be some sort of validation that `interval > 0` before the while loop?

brettimus:
can't quite figure out what's causing this, but i just ran into a strange spacing issue:

image

arendjr:
The spacing issue is the empty issue on the left, right? I suppose it can happen if you drag, but don't receive new data?

brettimus:
> 

ah no i think i can reproduce - happens when you look at a 12+ hour window i think (in explorer, it started getting funky after i switch to "Last 12 hours" and it remained funky for all other time ranges greater than that)

brettimus:
important context: i think it's because there isn't any data before 6 hours ago for the function i'm looking at. so, the absence of data leading up to when this function started to receive requests seems to be tripping up the gridlines

brettimus (changes requested):
so it seems the tickSuggestions need to handle the case where we only have a small slice of data across a wide time range. that's my understanding of what's messing up the chart i posted. (the buckets used in suggestions for the x-axis ticks are calculated based off of the observed time values, but it's possible that the timeRange.from is significantly earlier than the min value from the datapoints in the data set, e.g.)



otherwise, i'd say getting the grid lines to move when you're in the "shift + grab" scenario is a non-blocker

fiberplane-charts/src/Mondrian/generator/utils/attachSuggestionsToXAxis.ts
arendjr: Not a bad idea.

arendjr:
Should be all good now.

brettimus:
so i'm still getting the spacing issue locally, when there are no data points at all on one end of the timeRange 

image

this is easy to reproduce when you get something new started in explorer, and maybe you have a few minutes of actual data, but then your query asks for data in the past 15minutes. can pop up steps to repro if you want, or if you don't mind i can take a crack at it too?

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -61,31 +72,122 @@ export const GridWithAxes = memo(function GridWithAxes({
...
+  if (ticks.length < 2) {
+    return;
+  }
+
+  const interval = ticks[1] - ticks[0];
+  const scaleToAxis = (value: number) =>
+    scale((value - axis.minValue) / (axis.maxValue - axis.minValue));
+
+  let preTick = ticks[0] - interval;
+  while (ticks.length < maxTicks && scaleToAxis(preTick) > -0.1 * max) {
```
brettimus: i think this preTick shows up unexpectedly sometimes, especially when the viewport width is slightly less wide

image

fiberplane-charts/src/Mondrian/generator/utils/attachSuggestionsToXAxis.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (!firstBucketTimestamp || !lastBucketTimestamp) {
+    return;
+  }
+
+  if (interval <= 0) {
+    return;
+  }
+
+  const isAxisMinFarAwayFromFirstBucket =
+    firstBucketTimestamp - xAxis.minValue > 2 * interval;
```
brettimus: you might want to adjust this interval, but this was my basic heuristic to short-circuit suggestions when there are missing data on the left and right side of the data. it seems to work for now, since `getTicks` will fallback to just using the axis min and max

brettimus (approved):
i added a simple solution to address the case when the xAxis min/max are far away from the bucket min/max. you'll probably want to review that!

also think i ran into some funkiness with the `preTick` logic, but i don't consider that a blocker.

PR #73, opened 2023-07-13
[WIP] Add support for extra vertical lines in line charts

flenter:
# Description

This vertical lines can be used to show events on graphs (like when a deploy happened).

PR #76, opened 2023-07-19
`fiberplane-charts`: Add events to `<MetricsChart />`

arendjr:
# Description

This adds an `events` prop to `<MetricsChart />` which will be used to render vertical lines representing the events, with appropriate tooltips.

Supercedes #73.

# Checklist

* Explorer: https://github.com/autometrics-dev/explorer/pull/87


fiberplane-charts/src/MetricsChart/MetricsChart.tsx
```diff
@@ -44,48 +56,64 @@ const InnerMetricsChart = memo(function InnerMetricsChart(
...
     (
       anchor: TooltipAnchor,
-      [timeseries, metric]: [Timeseries, Metric],
+      [series, point]: [SeriesSource, Metric | ProviderEvent],
     ): CloseTooltipFn =>
       props.showTooltip?.(
         anchor,
-        <Tooltip timeseries={timeseries} metric={metric} />,
+        series.type === "events" ? (
+          <EventTooltip event={point as ProviderEvent} />
```
brettimus, line 113: might be overkill, but i'm wondering if the tooltip component could be generic, then render the proper contents based off of the type of the data passed to it? 

so the inside of `<Tooltip data={point}>` would have some sort of `isProviderEvent(data)` and `isTimeseriesPoint(data)` 


it'd have the benefit of avoiding `as` here, but would push some unsavory complexity into `<Tooltip>`

arendjr: Technically we could, but the main reason for the `as` is because I added a `type` property to the `series` prop, but I didn't do the same for the points. So `MetricsChart` "knows" that `series.type === "events"` means the points are `ProviderEvent`s and otherwise they are `Metric`s. But because `point` doesn't have a `type` field, detecting it inside `<Tooltip />` would also be tricky, or unsavory as you put it :)

The cleanest solution would be to give `point` a `type` field as well. Then we could have clean `isProviderEvent()` and `isTimeseriesPoint()` functions like you suggest. But it comes with a runtime cost, since all metrics need to mapped to new objects with a `type` field, and it would require adding a new type as well. So given those trade-offs, I felt this `as` was justified, since we can just determine the type based on the type of the series.

brettimus (approved):
i think we can merge this in! my guess is we'll want a design pass later this week on the vertical line for events, but all these changes work without introducing regressions (and even improve some bits as far as i can tell, like the y axis when rendering a single point)

PR #77, opened 2023-07-24
Add `fiberplane-prometheus-query` package

arendjr:
# Description

Creates a library for the TypeScript version of (a subset of) our Prometheus Provider.

I ended up calling it `fiberplane-prometheus-query` rather than `prometheus-provider-ts`, since it's unlikely to ever gain full feature parity with the actual provider, and it also doesn't return the `Blob` types like a real provider would.

Fixes https://github.com/autometrics-dev/explorer/issues/109

# Checklist


fiberplane-prometheus-query/src/providerTypes.ts
```diff
@@ -0,0 +1 @@
+../../fiberplane-provider-protocol/ts-runtime/types.ts
```
brettimus, line 1: is this a special syntax or does this need an import/export?

fiberplane-prometheus-query/src/timeseries.ts
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Fetches timeseries data from the Prometheus `query_range` endpoint.
+ *
+ * @param query Prometheus query string.
+ * @param timeRange Time range to fetch the data for.
+ */
+export async function querySeries(
+  query: string,
+  timeRange: TimeRange,
+  { baseUrl, headers }: QueryPrometheusOptions,
```
brettimus: i was thinking for future compatibility it could accept a fourth (optional) param of type `FetchArgs` — thoughts?

fiberplane-prometheus-query/src/timeseries.ts
```diff
@@ -0,0 +1,173 @@
...
+  return result.map(metricEntryToTimeseries);
+}
+
+/**
+ * Maps an entry from the result array returned by the Prometheus API to a
+ * Timeseries object.
+ *
+ * @param entry An entry returned by the Prometheus `query_range` API.
+ */
+function metricEntryToTimeseries(entry: unknown): Timeseries {
```
brettimus: nice - the typechecking here is a big improvement upon what's in dora

fiberplane-prometheus-query/src/timeseries.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  const amount = Math.ceil(step);
+  return [`${amount}${unit}`, amount * secondsMultiplier];
+}
+
+function getSecondsFromTimestamp(timestamp: Timestamp): number {
+  return +new Date(timestamp) / 1000;
+}
+
+function hasStringValues(
```
brettimus: maybe rename? `areAllValuesStrings` 

my reasoning is that this will return `true` for empty objects, and when i first read the function name above, i thought it might return false, since an empty object doesn't have any string values.

fiberplane-prometheus-query/src/timeseries.ts
brettimus: would you be okay with exporting this as well?

fiberplane-prometheus-query/src/timeseries.ts
```diff
@@ -0,0 +1,173 @@
...
+}
+
+/**
+ * Create a step size string from a time range, to be used in Prometheus queries
+ * This logic was translated from the Rust provider codebase
+ *
+ * @returns Step size, both as a string to use with Prometheus, and a number of
+ *          seconds.
+ */
+function getStepFromTimeRange(timeRange: TimeRange): [string, number] {
```
brettimus: how would you feel about exporting this utility too?

brettimus (approved):
approved just with one dangling question about the `providerTypes.ts` file

and if you're open to exporting the helpers too (`metricEntryToTimeseries`, `getStepFromTimeRange`) i'd be in favor of that

fiberplane-prometheus-query/src/providerTypes.ts
arendjr: It's a symlink, for which Git just stores the path it points to.

fiberplane-prometheus-query/src/timeseries.ts
arendjr: You mean `RequestInit`, I suppose? I can do that :)

fiberplane-prometheus-query/src/timeseries.ts
arendjr: sure!

fiberplane-prometheus-query/src/timeseries.ts
arendjr: no problem.

hatchan (approved):

fiberplane-prometheus-query/src/providerTypes.ts
brettimus: ooohh i've never seen a symlink in a pr before! hence the confusion 😂  - shoulda just looked at the file locally

PR #78, opened 2023-07-25
Fix chart clipping issues

brettimus:
# Description

This PR fixes two issues with clipping in the updated charts

1. Tick labels that were too close to the max of the axis would get clipped
2. Points or lines at the min or max of the chart would be clipped

## Tick Clipping

CoreChart/GridWithAxes was modified to use a helper function called `removeLastTickIfTooCloseToMax`, which will remove the final tick if it is within a certain distance of the axis max.

For the x-axis I settled on a heuristic for the max tick value of 1/2 the tick interval.

For the y-axis, I settled on a heuristic for the max tick value of 1/3 of the tick interval.

I chose different heuristics here, since labels are typically wider than they are tall, so more likely to clip along the x-axis.

I know this isn't a perfect solution, but it worked for the charts I was testing on.

## Point Clipping

For the point clipping, I did two things:

First, I simple extended the clip path for the chart contents so it was 2 units higher and 2 units lower. This is because our points have a max radius of 2 (when they are emphasized).

**Before**:

image

**After**:

image

### Part II: Fixing Spark Chart Clipping

Second, the spark charts still had clipping issues after implementing the above. In this case, their clip path was actually higher than the svg element itself. So, I modified the height of the SVG to be the `max` of the svg-height and clip path height. This resolved the issue.

**Before**:

image

After

image

# Checklist

~- [ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~


fiberplane-charts/src/CoreChart/constants.ts
```diff
@@ -1,3 +1,8 @@
+import { POINT_RADIUS_FOCUSED } from "./ChartShape";
+
 // Dimensions.
 export const HEIGHT = 275;
 export const MARGINS = { top: 0, bottom: 20, left: 38, right: 0 };
+// If a point is directly on the edge of the chart, it can be cut off.
+// The overflow margin ensures that the point is still visible.
+export const CHART_SHAPE_OVERFLOW_MARGIN = POINT_RADIUS_FOCUSED;
```
brettimus, line 11: i have a feeling you might not like this, BUT! 

since we might end up increasing the point size at some point (no pun intended), this seemed to be the clearest way to communicate in code that the chart clip path margins are determined by the size of the points on the graph

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -222,3 +225,58 @@ function useCustomSpring(value: number) {
...
+ * Returns a maximum allowed tick value for the x-axis.
+ *
+ * Heuristic:
+ *   If a tick's distance to the maxValue is within 1/2 the size of the tick-interval,
+ *   the tick will get dropped.
+ *
+ * Note that the heuristic was determined by trial and error.
+ */
+const getMaxXTickValue = (ticks: Array<number>, maxValue: number) => {
+  const interval = ticks[1] - ticks[0];
```
arendjr, line 270: It's best to verify the length of the `ticks` first.

I notice there's a check in the caller, but that's easily overlooked during a refactor, and it makes it harder to reason about the code if you first have to understand/verify the call sites.

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -222,3 +225,58 @@ function useCustomSpring(value: number) {
...
+ * Returns a maximum allowed tick value for the x-axis.
+ *
+ * Heuristic:
+ *   If a tick's distance to the maxValue is within 1/3 the size of the tick-interval,
+ *   the tick will get dropped.
+ *
+ * Note that the heuristic was determined by trial and error.
+ */
+const getMaxYTickValue = (ticks: Array<number>, maxValue: number) => {
+  const interval = ticks[1] - ticks[0];
```
arendjr, line 288: It's best to verify the length of the `ticks` first.

fiberplane-charts/src/CoreChart/constants.ts
```diff
@@ -1,3 +1,8 @@
+import { POINT_RADIUS_FOCUSED } from "./ChartShape";
```
arendjr: It's better to just move the `POINT_FOCUS` constants to this file, rather than letting `constants.ts` depend on the shapes.

fiberplane-charts/src/CoreChart/constants.ts
arendjr: Au contraire, I like these semantic clarifications :)

arendjr (approved):
Nice one!

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
brettimus: very true — it's done!

PR #84, opened 2023-07-31
Add configuration to not render gradients for AreaShape and LineShape

brettimus:
# Description

This PR adds an option to `CoreChart` and by extension `MetricChart`, which can be used to remove gradients from Line and Area shapes when they are rendered.

I added the prop `areaGradientShown` to the `CommonShapeProps` type, which is passed down from the `CoreChart`.

For backwards compatibility, this defaults to `true`. 

I also added this prop to `SparkChart`, so we can turn off gradients in the Spark charts

Here's why I made the change:

**Before**

image

**After**

image

# Checklist

- ~[ ] The OpenAPI schema and generated client have been updated.~
- ~[ ] New models module has been added to api generator xtask~
- ~[ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~
- ~[ ] The CHANGELOG is updated.~


arendjr (commented):
Please just add the prop to the `CommonShapeProps` that we also use for passing down which shape is selected. Then we don’t need to add another context.

arendjr (approved):

PR #86, opened 2023-08-04
Fix issue with ticks when your data only covers a small slice of the axis range

brettimus:
Fixes https://github.com/autometrics-dev/explorer/issues/136

# Description

This PR supplies a fallback when `getTicksFromSuggestions` pares down the suggested ticks to a list of length 1. 

I initially tried to instead pass a fallback interval to `extendTicksToFitAxis`, which could be used when the `ticks` argument only had one element. However, this was only a partial fix, as `extendTicksToFixAxis` would simply add "preTicks" until it reached the maximum number of allowed ticks. We still ended up with a large gap in some cases.

# Checklist


fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -213,12 +208,14 @@ function extendTicksToFitAxis(
   max: number,
   scale: Scale,
   maxTicks: number,
+  fallbackInterval?: number,
```
arendjr: I don't think this is even a "fallback" interval, it's **the** interval :) So this function doesn't even need to determine it on its own anymore. Meaning it can also drop the early return; we just shouldn't call it if there is no interval.

arendjr (approved):

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
brettimus: fixed!

PR #87, opened 2023-08-04
Stylize the vertical event lines with a dash

brettimus:
# Description

This renders the event vertical lines as a dashed line. 

I ended up passing the `shapeList` source type down to the shapes to accomplish this. 

The problem is that `<ChartContent>` uses generics for its props, even though in practice its `S` (source) always has a `type` property. 

Because of this, I had to write a kinda hacky helper `getSourceTypeIfExists` to inject the source type as a prop on the shapes.

I was thinking of making `S` extend a base `type` that made sure the source always had a `.type` property. However, this would've been a very invasive change, so I opted for the solution you see before you.

# Checklist


brettimus:
sorry i worked off of the old branch and it made the history crazy 🌀

arendjr:
Unfortunately this logic would be impossible to port to the Rust version of the renderer, since the `type` property doesn’t exist there. I think a cleaner solution would be to add the `strokeDasharray` field to the `Line` shape.

Happy to pick this up 👍

brettimus:
> Unfortunately this logic would be impossible to port to the Rust version of the renderer, since the `type` property doesn’t exist there. I think a cleaner solution would be to add the `strokeDasharray` field to the `Line` shape.
> 
> Happy to pick this up 👍

that'd be great if you could! i'll close this PR then

PR #89, opened 2023-08-08
Mondrian exporter to SVG and PNG

arendjr:
# Description

This PR extends the Rust version of Mondrian to allow exporting to SVG and PNG (and other bitmap formats supported by the `image` crate). This contains some major changes for the Mondrian crate:

* Tick formatters had to be implemented in Rust too, and for consistency I've exposed the same implementation in TypeScript.
* Font handling had to be added. This is handled by the `fontdb` crate, but the caller is supposed to initialize the database and pass it along. This works well with our API, where we can just use `include_bytes!()` for the font we need.
* I've also added feature flags so people can disable the image generation or the dependency on Fiberplane models, if they prefer.
* There's a **breaking change** in the `fiberplane-charts` library, because I've changed the format for time values from milliseconds to seconds. It was quite confusing the X axis values used milliseconds, while other time values use seconds, which became especially relevant when implementing the tick formatters.

# Checklist


fiberplane-charts/src/tickFormatters.test.ts
```diff
@@ -0,0 +1,43 @@
+import { getScientificFormatterForAxis } from "./tickFormatters";
+
+test("it can format with scientific notation", () => {
```
brettimus: this test file is clearer about behavior than anything i saw for the d3 value formatters

brettimus (approved):

PR #96, opened 2023-08-14
Publish `mondrian-charts` and `@fiberplane/fiberplane-charts`

arendjr:
# Description

This sets up `mondrian-charts` for publication to Crates.io and `@fiberplane/fiberplane-charts` and `@fiberplane/prometheus-query` for publication to NPM.

The main changes are a lot of documentation added and the removal of the `dist` folder from `fiberplane-charts`. We will have to see if the latter really makes for a more pleasant dev experience (I know nobody likes regenerating the `dist` folder, but requiring releases before we can use new features downstream *might* be even worse?).

Also contains some smaller tweaks and a bit of API streamlining that I ran into as I was documenting things. This also introduces a few backwards incompatibilities again, but fortunately nobody depends yet on the Rust crate, and for the TypeScript package it was inevitable due to the name change (added the `@fiberplane` namespace).

# Checklist

- :warning: **The changes are not backwards compatible.**


brettimus:
regarding the local dev experience for charts:

it'd be good to document how we expect people to develop locally, because we're still going to be doing that quite a bit. 

i think it'd still be possible to generate a build of the charts and link to it locally, which is how i've been doing things anyhow most of the time. that said, it is a lot of _steps_:

- run `yarn build` in fiberplane-charts
- run `yarn pack --out charts-zzz.tgz`
- modify package.json in e.g., dora to point fiberplane-charts to the `charts-zzz.tgz` file

maybe there's a way to make this more seamless? like a script that wraps up all three steps?

fiberplane-charts/.npmignore
```diff
@@ -0,0 +1,7 @@
+node_modules/
```
brettimus: when `.npmignore` is present, `.gitignore` is not used at all. for safety, we might wanna add `.vscode` to this as well, or any other local files that might find their way into the published package (granted, if we build + publish from ci, this is less of a concern)

fiberplane-charts/README.md
```diff
@@ -4,22 +4,38 @@
...
+    "@fiberplane/fiberplane-charts": "https://git@github.com/fiberplane/fiberplane.git#workspace=fiberplane-charts&head=<branch_name>"
+  }
 ```
+
+Run `yarn` to install as usual.
+
+If you want to update `fiberplane-charts` while using the above syntax, you may
+use `yarn up "@fiberplane/fiberplane-charts@<full-dependency-url>"`.
+
+Make sure to remove the `dist/` folder before opening a PR.
```
brettimus: this relates to [my comment about the local dev experience](https://github.com/fiberplane/fiberplane/pull/96#issuecomment-1683578058)

this upside of what you've written here is it makes the temporary build of charts easier to share across multiple people working on the same PR.

however, manually removing `dist` does not seem ideal.

the local dev workflow i outlined has the benefit that `dist` stays out of git; however, if you wanted someone to check your work, then they would have to repeat the process of creating a local build and linking to it

curious about your thoughts

fiberplane-charts/.npmignore
arendjr: Good call!

fiberplane-charts/README.md
arendjr: Manually removing `dist` is not ideal indeed, but at least I think it's safe enough now that CI will fail if you check it in. So you *can* check it in if you want to share it with someone to let them try it out, but CI will complain so you wouldn't merge it into `main` at least.

But I agree your `yarn pack` workflow should be the one we recommend by default.

brettimus (approved):

mellowagain (approved):

PR #98, opened 2023-08-18
Configure whether or not to animate the y scales

brettimus:
# Description

This PR tackles https://github.com/autometrics-dev/explorer/issues/194 — there have been multiple comments about the awkwardness of the initial Y-scale animation, so I think we should leave it up to consumers of the charting library to specify when they want that animation to occur. 

So, to that effect, I've added a prop called `shouldAnimateYScale`. This is ultimately passed down to the `useCustomSpring` hook inside the `GridWithAxes` component, which allows it to skip animating the value of axes. 

As far as I can tell, this allows users of the library to determine _when_ scale animation should happen (e.g., do not animate if this is our first time showing data, but _do_ animate when we've change the data we're showing)

# Checklist


fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -251,9 +253,14 @@ function useCustomSpring(value: number) {
...
-    const controls = animate(motionValue, value, spring);
-    return controls.stop;
-  }, [motionValue, value]);
+    if (shouldAnimate) {
+      const controls = animate(motionValue, value, spring);
+      return controls.stop;
+    }
+
+    setCurrent(value);
+    return () => {};
```
arendjr: I think we even have a `noop` util in there :)

arendjr (approved):

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
brettimus: will add!

PR #100, opened 2023-08-19
Cover a bunch of edge cases with ticks for uniform value graphs

brettimus:
This is an alternative solution for improving y-ticks on graphs of uniform values

Uniform error rate of 0%:

image

Uniform latency of .00475s

image

Some 100% success rates:

image


fiberplane-charts/src/Mondrian/generator/utils/getYAxisForConstantValue.ts
```diff
@@ -26,6 +40,47 @@ export function getYAxisForConstantValue(value: number): Axis {
...
+  axis.tickSuggestions = [];
+
+  let currentTick = firstTick;
+
+  while (currentTick <= axis.maxValue) {
+    axis.tickSuggestions.push(currentTick);
+    currentTick += interval;
+  }
+
+  return;
```
arendjr: This `return` seems a bit superfluous :)

arendjr (approved):
Very nice! :+1:

arendjr:
Btw, I see the PR is against my PR, so if you merge into that one I can update the Rust version if you like.

brettimus:
> Btw

yeah! i'll tidy this up then merge into yours

PR #105, opened 2023-09-20
Rename xtask

arendjr:
# Description

This is a minor change that allows the `xtask` crate from this repo to co-exist with the crates from our other repos when combined using Git submodules.


arendjr:
Closing in favor of #106.

PR #107, opened 2023-10-05
Bump prometheus query package

laulauland:
The currently published package `@fiberplane/prometheus-query` is empty (doesn't include a `dist/` folder). This bumps the prometheus-query version so a fix for an empty package can be published.


flenter:
I'm fine with bumping the version. But the publish flow is largely based on what you have locally and this PR doesn't seem to fix it

flenter (approved):

arendjr (approved):

PR #114, opened 2023-10-23
Rename @fiberplane/fiberplane-charts

flenter:
# Description

This PR renames `@fiberplane/fiberplane-charts` to just `@fiberplane/charts`. This is a breaking change though the two main products that use it from us haven't really been switched over yet to the new setup so it's not a major breaking change in that sense.

The thing we may want to think about is what to do with the old @fiberplan/fiberplane-charts package on npm. It would be nice to publish something that says people should use the new `@fiberplane/charts` instead.

### DIFF
--- a/fiberplane-charts/package.json
+++ b/fiberplane-charts/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "@fiberplane/fiberplane-charts",
+    "name": "@fiberplane/charts",
     "version": "0.2.0",
     "description": "Charts for visualizing metrics",
     "author": "Fiberplane <info@fiberplane.com>",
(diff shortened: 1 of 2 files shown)


arendjr:
You can mark packages as obsolete on NPM, though that doesn't come with a notice that says *why* something is deprecated. I would still do that though, but the best is to also publish a new version with only an updated README that says where to find the new package.

arendjr (approved):

mellowagain:
you can mark packages as deprecated with a custom message: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

arendjr:
The docs say that, yeah, but I've tried before and got no such option :shrug: Happy to hear if it works though :sweat_smile:

arendjr:
Oh wait, it's a CLI-only feature. I might have missed that!

flenter:
I guess this PR is no longer relevant as #115 already contains the changes

PR #118, opened 2023-11-08
TypeScript magic strikes again

arendjr:
This resolves the issue of needing to specify Prettier and SVGO dependencies in the workspace, despite not using those dependencies.

* I got this [solution from StackOverflow](https://stackoverflow.com/questions/72734737/tsconfig-json-shows-error-entry-point-for-implicit-type-library-glob), so don't ask me exactly why this works, but I *think* it works because without it TypeScript will try to pull in type definitions of transitive dependencies that mess with our builds. But this only worked for fixing the builds, not for the type-checking of the project.
* To get the type-checking to work, I also had to exclude the `rollup.config.ts` from the project config (by only explicitly including `src/**/*`) because otherwise some transitive dependencies of `@svgr/rollup` still got pulled in.

I also did some deduplication of packages.


arendjr:
~~Ah wait, so it fixed the builds, but broke the type checks. Of course.~~ **Fixed:** see updated description.

flenter (approved):

hatchan (approved):

oscarvz (approved):

PR #126, opened 2023-11-22
Add Front Matter Schema models to workspaces, and notebooks

gagbo:
- Add front matter schema to models
- Add front matter schema model to workspaces
- Add front matter schemas to notebook and templates

# Description

Please include a summary of the change and which issue is fixed. Please also
include relevant motivation and context.

Fixes AM-178 AM-163 AM-166 AM-168

# Checklist


stephlow (approved):

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,181 @@
...
+)]
+#[non_exhaustive]
+pub struct FrontMatterStringSchema {
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub icon_name: Option<String>,
+    // Skip serialization if the bool is false, and defaults to false, and the setter in typed_builder will set the field to true.
+    #[builder(setter(strip_bool))]
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub multiple: bool,
```
flenter, line 217: Can you make this optional? (also applies elsewhere)

flenter: Oh I guess it won't be included when its serialised. Nevermind!

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub options: Option<Vec<FrontMatterEnumNumberValue>>,
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub default_value: Option<FrontMatterEnumNumberValue>,
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub prefix: Option<String>,
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub suffix: Option<String>,
```
arendjr, line 194: I feel these struct definitions could do with some newlines between fields... these are some *very* crowded definitions :sweat_smile:

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,270 @@
...
+#[non_exhaustive]
+pub struct FrontMatterNumberSchema {
+    #[builder(default, setter(into))]
+    pub display_name: String,
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub icon_name: Option<String>,
+    // Skip serialization if the bool is false, and defaults to false, and the setter in typed_builder will set the field to true.
+    #[builder(setter(strip_bool))]
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
```
arendjr, line 161: I like the "trick" of using `std::ops::Not::not` for this :+1:

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub multiple: bool,
+    pub allow_extra_values: bool,
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub options: Option<Vec<FrontMatterEnumNumberValue>>,
```
arendjr, line 182: It might be more convenient to type it like this:

```suggestion
    #[builder(default, setter(into))]
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub options: Vec<FrontMatterEnumNumberValue>,
```

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,270 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub options: Option<Vec<FrontMatterEnumBase64UuidValue>>,
+    #[builder(default, setter(into, strip_option))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub default_value: Option<FrontMatterEnumBase64UuidValue>,
+}
+
+// NOTE: The cleaner way would be to have a generic type FrontMatterEnumValue<T>,
+// but it's impossible to _conditionnally_ add the `Serializable` trait bound on
+// the inner type T only when there is the "fp-bindgen" feature.
```
arendjr, line 340: While you can't conditionally add the `Serializable` trait, wouldn't it make sense to define the type to only apply to types that _are_ serializable? I.e. `FrontMatterEnumValue<T: Serializable>`

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// NOTE: The reason those are struct (new types) instead of "just" being the
```
arendjr: Nit: the "new type" idiom refers to a struct with a single nameless field, which doesn't apply here.

fiberplane-models/src/notebooks/mod.rs
```diff
@@ -361,15 +371,27 @@ pub struct NewTemplate {
...
+        name: Name,
+        description: impl Into<String>,
+        front_matter_schema_names: &[impl ToString],
+        body: impl Into<String>,
+    ) -> Self {
         Self {
             name,
             description: description.into(),
+            front_matter_schema_names: front_matter_schema_names
+                .into_iter()
```
arendjr: Clippy.

fiberplane-models/src/notebooks/operations.rs
```diff
@@ -425,6 +432,86 @@ pub struct ClearFrontMatterOperation {
     pub front_matter: FrontMatter,
 }
 
+/// Adds a front matter key in a notebook (at the end)
```
arendjr: I think I'd call this `AppendFrontMatterSchemaOperation` instead. "Append", since its semantics define it to only add at the end anyway. And "Schema", since there seems to be some ambiguity around operations called `...KeyOperation`: Sometimes the key appears to refer to the schema definitions, and sometimes it's the key in the value map.

fiberplane-models/src/workspaces.rs
```diff
@@ -41,6 +43,9 @@ pub struct Workspace {
 
     #[builder(setter(into))]
     pub updated_at: Timestamp,
+
+    #[builder(default)]
+    pub front_matter_schemas: HashMap<String, FrontMatterSchema>,
```
arendjr: Please use a `BTreeMap` here so we'll have consistent serialization.

arendjr (commented):
Nice job!

I do think we'd better wait with merging until the operations are supported and tested in `fiberplane-ot` so we know the definitions of the new operations are sufficient. Especially `AppendFrontMatterSchemaOperation` will be slightly tricky since it subtly conflicts with operations of its own kind (they can't be replayed in arbitrary order).

For `MoveFrontMatterSchemaOperation` I'd advise to carefully look into how we implement this for moving cells.

fiberplane-models/src/workspaces.rs
hatchan: What do you mean with "consistent serialization"? In the case of JSON both will be serialized as a object, right?

fiberplane-models/src/front_matter_schemas.rs
hatchan: Would we also need `default` in case the field is not present, or empty? So something like:
```
#[serde(default, skip_serializing_if = "Vec::is_empty")]
```

fiberplane-models/src/workspaces.rs
arendjr: Yes, but the order in which keys are serialized is undefined with a hashmap, while it's fixed for a btree map. This can be slightly annoying with integration tests, for instance, because the string representation is no longer defined. JS/TS also have a fixed iteration order for objects, but if you parse incoming data, that order is dependent on the order in which it was serialized, so it can result in subtle differences down the line.

fiberplane-models/src/front_matter_schemas.rs
arendjr: Yes, you're right. Updated my suggestion.

fiberplane-models/src/front_matter_schemas.rs
arendjr: Btw, I think I ran into the same issue as you in my PR and I'm solving it by adding the following to `fiberplane-models/src/lib.rs`:

```rs
#[cfg(feature = "fp-bindgen")]
pub trait MaybeSerializable: Serializable {}

#[cfg(not(feature = "fp-bindgen"))]
pub trait MaybeSerializable {}
```

Then you can just add a trait bound for `MaybeSerializable`, which should solve your problem too.

arendjr: See: https://github.com/fiberplane/fiberplane/pull/131/commits/0ea57a5510c443add49398ce090871a6e3f110ee

arendjr: One downside: You may also need to add empty blocks like this: `impl MaybeSerializable for AutometricsFunction {}`

As an alternative solution, you can also use this: https://stackoverflow.com/a/68416114/1496305

fiberplane-models/src/front_matter_schemas.rs
gagbo: Sure, will do :)

fiberplane-models/src/front_matter_schemas.rs
gagbo: The issue with using a Vec here is that we lose the information about whether the field is:
- a free form field, or
- a field where you should create "tags" to attach as the value of the front matter key, but without pre-determined options

For numbers that might not be a huge deal but it’s important for strings

fiberplane-models/src/front_matter_schemas.rs
gagbo: I thought about doing this afterwards, maybe the trait should be a top-level helper in the models

gagbo:
> Especially AppendFrontMatterSchemaOperation will be slightly tricky since it subtly conflicts with operations of its own kind (they can't be replayed in arbitrary order).

I don’t really know yet how to avoid order of operations issues when we want to insert things in a vector in the end. I might try adding the "current front matter length" in the operation and the change, but I guess I'll see when I implement the operation application in OT

fiberplane-models/src/workspaces.rs
gagbo: Does that mean the Front Matter in notebooks should change too? If we have concurrent edits of Front Matter entries, the `serde_json::Map` doesn't enforce any order either

fiberplane-models/src/front_matter_schemas.rs
arendjr: Fair enough!

It does highlight these fields could do with some doc comments, because I had not inferred those meanings from the types alone.

fiberplane-models/src/front_matter_schemas.rs
arendjr: Yeah, I agree it should be a reusable helper, so I put it in `lib.rs` in my PR. As I mentioned to @hatchan the other day, we're also getting to the point that I think we should just have something like `FiberplaneStruct` and `FiberplaneEnum` derive macros that take all our conventions into account. But that might be more for a hackday or something :)

fiberplane-models/src/workspaces.rs
arendjr: The docs say `serde_json::Map` is backed by a `BTreeMap` as well, so I think that one is already fine: https://docs.rs/serde_json/latest/serde_json/map/index.html

arendjr:
> I don’t really know yet how to avoid order of operations issues when we want to insert things in a vector in the end. I might try adding the "current front matter length" in the operation and the change, but I guess I'll see when I implement the operation application in OT

To be honest, I'm also hesitant saying with too much certainty how to resolve the conflicts, because it wouldn't be the first time an issue doesn't come to light until I actually try to implement the OT logic either :) But I *think* you should be able to resolve the conflicts without need for transformation here. But if that's not the case, an `InsertFrontMatterSchemaOperation` with an `index` would indeed be a fine alternative. Then it could even be used for insertion at other indices as well. But that one will definitely require more transformations to be implemented.

fiberplane-models/src/front_matter_schemas.rs
arendjr: Would you mind added some docs still?

fiberplane-models/src/lib.rs
```diff
@@ -114,3 +116,16 @@ impl<'de> Visitor<'de> for U32Visitor {
...
+
+/// Helper trait to use as a trait bound on generic types that must be serializable when
+/// the `fp-bindgen` feature is enabled
+pub trait BindgenSerializable {}
+
+#[cfg(feature = "fp-bindgen")]
+impl<T> BindgenSerializable for T where
+    T: fp_bindgen::prelude::Serializable + for<'de> de::Deserialize<'de>
+{
+}
```
arendjr, line 128: Nice, I hadn't thought of the auto-impls with a generic yet :+1:

fiberplane-openapi-rust-gen/src/routes.rs
```diff
@@ -34,6 +34,7 @@ pub(crate) fn generate_routes(
 
     let mut writer = BufWriter::new(file);
 
+    writeln!(writer, "#![allow(clippy::too_many_arguments)]")?;
```
arendjr, line 37: Nice one!

arendjr (approved):
Really nice job! Please wait with merging until the `monofiber` PR is ready too.

fiberplane-models/src/lib.rs
gagbo: I don’t really use this one yet since I ended up not using generics, but I think it’s going to be the cleanest way to have generics with the Serializable trait bound.

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,316 @@
...
+/// The order of the elements in the schema drives the order of
+/// rendering elements.
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize, Default)]
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::front_matter_schemas")
+)]
+#[repr(transparent)]
+pub struct FrontMatterSchema(pub Vec<FrontMatterSchemaEntry>);
```
arendjr, line 62: I think we may want to provide an `impl Deref<Vec<FrontMatterSchemaEntry>> for FrontMatterSchema`. I noticed the `.0` we need to do in the consumer sites, and it makes for a not very readable API.

fiberplane-models/src/front_matter_schemas.rs
```diff
@@ -0,0 +1,316 @@
...
+pub struct SerializableEqFloat(OrderedFloat<f64>);
+
+#[cfg(feature = "fp-bindgen")]
+impl Serializable for SerializableEqFloat {
+    fn ident() -> fp_bindgen::types::TypeIdent {
+        fp_bindgen::types::TypeIdent::from("f64")
+    }
+
+    fn ty() -> fp_bindgen::types::Type {
+        fp_bindgen::types::Type::Primitive(fp_bindgen::primitives::Primitive::F64)
```
arendjr: I think this *should* help so you don't need the extra patch in the FiberKit bindings:

```suggestion
        fp_bindgen::types::Type::Custom(fp_bindgen::types::CustomType {
            fp_bindgen::types::TypeIdent::from("SerializableEqFloat"),
            rs_ty: "f64".to_owned(),
            ts_ty: "number".to_owned(),
            ..Default::default()
        })
```

I'll admit it would've been nice if `fp-bindgen` auto-detected that you're mapping to a primitive and would've caught that for you :)

gagbo: I tried but I don’t think it changed anything (the patch applied cleanly the same way), maybe it was a caching issue I’ll keep an eye on it

PR #135, opened 2023-12-01
Add configurable stroke width to mondrian-charts

actualwitch:
# Description

This pr adds `shape_stroke_width` optional field to `ChartOptions` with a default of 1 (current value). This change will be used by our new alert slack bot that you should definitely check out 👀

# Checklist

* Monofiber: https://github.com/fiberplane/monofiber/pull/163

### DIFF
--- a/mondrian-charts/src/chart_to_svg/constants.rs
+++ b/mondrian-charts/src/chart_to_svg/constants.rs
@@ -12,6 +12,9 @@ pub(super) const POINT_RADIUS_FOCUSED: u16 = 4;
 // This overflow margin ensures that the point is still visible.
 pub(super) const CHART_SHAPE_OVERFLOW_MARGIN: u16 = POINT_RADIUS_FOCUSED;
 
+// Stroke width of chart content lines
+pub(super) const SHAPE_STROKE_WIDTH: f32 = 1.0;
+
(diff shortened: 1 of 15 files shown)


mondrian-charts/src/chart_to_svg/constants.rs
```diff
@@ -12,6 +12,9 @@ pub(super) const POINT_RADIUS_FOCUSED: u16 = 4;
 // This overflow margin ensures that the point is still visible.
 pub(super) const CHART_SHAPE_OVERFLOW_MARGIN: u16 = POINT_RADIUS_FOCUSED;
 
+// Stroke width of chart content lines
+pub(super) const SHAPE_STROKE_WIDTH: i8 = 1;
```
arendjr: Would it ever make sense to allow negative values? :) If not, I think `u8` would be preferred, although I'd also be inclined to just use `f32` so that fractions can be specified, which is valid in SVG too.

arendjr (commented):
Nice one! Would you like to update the TypeScript version as well?

mondrian-charts/src/chart_to_svg/constants.rs
actualwitch: didn't know fractions were valid here, fixed!

arendjr (approved):

PR #142, opened 2023-12-27
add integrations listing for user

mellowagain:
# Description

adds models for integration listing

fixes FP-3386

# Checklist

* Monofiber: https://github.com/fiberplane/monofiber/pull/193

### DIFF
--- a/fiberplane-models/src/integrations.rs
+++ b/fiberplane-models/src/integrations.rs
@@ -37,7 +37,7 @@ pub struct Integration {
 #[non_exhaustive]
 #[serde(rename_all = "lowercase")]
 pub enum IntegrationId {
-    GitHub
+    GitHub,
 }
 
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize, Display, EnumIter)]
--- a/fiberplane-models/src/integrations.rs
+++ b/fiberplane-models/src/integrations.rs
@@ -13,7 +13,7 @@ use typed_builder::TypedBuilder;
 )]
 #[non_exhaustive]
 #[serde(rename_all = "camelCase")]
-pub struct Integration {
+pub struct IntegrationSummary {
     pub id: IntegrationId,
     pub status: IntegrationStatus,
 
@@ -21,10 +21,6 @@ pub struct Integration {
     pub created_at: Timestamp,
     #[builder(setter(into))]
     pub updated_at: Timestamp,
-
-    #[builder(default, setter(strip_option, into))]
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub data: Option<serde_json::Value>,
 }
 
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Eq, Serialize, Display, EnumIter)]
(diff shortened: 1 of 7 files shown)


fiberplane-models/src/integrations.rs
```diff
@@ -0,0 +1,59 @@
...
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::integrations")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
+pub enum IntegrationStatus {
+    Connected,
+    Disconnected,
+    AttentionRequired { reason: String }, // todo: check if this gets serialized correctly
```
hatchan: I think I did ran into an issue _somewhere_ using this approach... Not sure what or where anymore, so not sure how useful this is haha.

fiberplane-models/src/integrations.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize, Display, EnumIter)]
+#[cfg_attr(
+    feature = "fp-bindgen",
+    derive(Serializable),
+    fp(rust_module = "fiberplane_models::integrations")
+)]
+#[non_exhaustive]
+#[serde(rename_all = "camelCase")]
```
hatchan: Shouldn't this be:

```suggestion
#[serde(rename_all = "snake_case", tag = "type")]
```

As that is what we're using in different places and we are including data.

schemas/openapi_v1.yml
```diff
@@ -2437,6 +2437,33 @@ components:
...
+        id:
+          type: string
+          enum:
+            - "TODO"
+        status:
+          type: string
+          enum:
+            - "connected"
+            - "disconnected"
+            - "attentionRequired"
```
hatchan: Don't think this will work with the serialization that we're using above (ie. with the embedded `reason`)

fiberplane-models/src/integrations.rs
arendjr: As long as you include the `tag = "type"` as @hatchan suggested in the previous comment, I think it should work.

hatchan (approved):

fiberplane-models/src/integrations.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub status: IntegrationStatus,
+
+    #[builder(setter(into))]
+    pub created_at: Timestamp,
+    #[builder(setter(into))]
+    pub updated_at: Timestamp,
+
+    #[builder(default, setter(strip_option, into))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub data: Option<serde_json::Value>,
```
stephlow:
```suggestion
    pub data: Option<Value>,
```

We need to import value at the top level otherwise fp-bindgen will generate this json:
```
    data?: serde_json::Value;
```

Which should be:

```
    data?: any;
```

stephlow (changes requested):

fiberplane-models/src/integrations.rs
mellowagain: nice catch! but i yeeted `data` off the struct anyways so should be good now

stephlow (approved):

hatchan (approved):

PR #147, opened 2024-01-05
Add updated_at to SubscriberChangedFocusMessage

stephlow:
# Description

In order to be able to de duplicate subscriber sessions with the same user id, we expose the `updated_at` property on the session so clients can select the last active session.

Fixes FP-3370

# Checklist

https://github.com/fiberplane/monofiber/pull/196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ major version hasn't changed.
   controlling the display and the properties of the front matter.
   Each workspace has a list of (editable) associated front matter schemas pickable when
   creating notebooks.
+- `fiberplane-models`: Add `updated_at` to `SubscriberChangedFocusMessage`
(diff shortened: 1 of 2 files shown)


hatchan (approved):

hatchan:
Actually do we need to make this optional, to satisfy backwards compatibility?

arendjr (approved):

arendjr:
> Actually do we need to make this optional, to satisfy backwards compatibility?

I don't think it's necessary, since these messages are sent by the server only. Older Studio versions will just ignore the added property, while newer ones will be able to rely on its presence.

stephlow:
> > Actually do we need to make this optional, to satisfy backwards compatibility?
> 
> I don't think it's necessary, since these messages are sent by the server only. Older Studio versions will just ignore the added property, while newer ones will be able to rely on its presence.

I double checked this with an older studio build and it seems to work 👍

PR #153, opened 2024-01-16
Add 'by name' support to mention front matter schema on notebook creation

gagbo:
# Description

Add support for mentioning front matter schemas by name when creating notebooks.
This is a step towards getting templates to expand with arbitrary front matter schemas.

Fixes FP-3458

# Checklist

- [x] ~~New models module has been added to api generator xtask~~

### DIFF
--- a/fiberplane-models/src/notebooks/mod.rs
+++ b/fiberplane-models/src/notebooks/mod.rs
@@ -114,20 +114,20 @@ pub struct NewNotebook {
     /// The "inline" front matter schema for the notebook.
     ///
     /// It will always be expanded first and have priority over anything defined in
-    /// `front_matter_schema_by_names`. The keys in common will be ignored.
+    /// `front_matter_collections`. The keys in common will be ignored.
     #[builder(default)]
     #[serde(default)]
     pub front_matter_schema: FrontMatterSchema,
 
     /// A list of front matter schema names that exist in the target workspace.
     ///
-    /// If `front_matter_schema_by_names` and `front_matter_schema` are both mentioned, then:
+    /// If `front_matter_collections` and `front_matter_schema` are both mentioned, then:
     /// - the `front_matter_schema` will be the first elements of the notebook front matter,
     /// - and keys from the named schema that already exist in `front_matter_schema` (or an earlier
     ///   entry in the list) will be ignored.
     #[builder(default)]
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
-    pub front_matter_schema_names: Vec<String>,
+    pub front_matter_collections: Vec<Name>,
 }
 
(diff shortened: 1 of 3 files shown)


fiberplane-models/src/notebooks/mod.rs
```diff
@@ -111,9 +111,23 @@ pub struct NewNotebook {
...
+
+    /// A list of front matter schema names that exist in the target workspace.
+    ///
+    /// If `front_matter_schema_by_names` and `front_matter_schema` are both mentioned, then:
+    /// - the `front_matter_schema` will be the first elements of the notebook front matter,
+    /// - and keys from the named schema that already exist in `front_matter_schema` (or an earlier
+    ///   entry in the list) will be ignored.
+    #[builder(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub front_matter_schema_by_names: Vec<String>,
```
arendjr: I would prefer to call this `front_matter_schema_names` like the old field. When something is called "by name" it sounds like you're looking something up, so I'd expect a `Map` type or something similar.

gagbo: Fair enough

fiberplane-models/src/notebooks/mod.rs
```diff
@@ -111,9 +111,23 @@ pub struct NewNotebook {
...
+
+    /// A list of front matter schema names that exist in the target workspace.
+    ///
+    /// If `front_matter_schema_by_names` and `front_matter_schema` are both mentioned, then:
+    /// - the `front_matter_schema` will be the first elements of the notebook front matter,
+    /// - and keys from the named schema that already exist in `front_matter_schema` (or an earlier
+    ///   entry in the list) will be ignored.
+    #[builder(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
+    pub front_matter_schema_names: Vec<String>,
```
hatchan: I think we should maybe call this `front_matter_schema_collections` as it is referring to 1 or more collections of schemas. It will also differentiate from the inline `front_matter_schema` more. 

Also I opted to not include the `_names` as I think it might be a bit too verbose and it is clear that it will be a reference.

hatchan: Another thing is that we probably want to use the `Name` struct. Since this collection will be referenced on the API, it should also get the same requirements/restrictions.

hatchan: https://github.com/fiberplane/fiberplane/blob/main/fiberplane-models/src/names.rs

gagbo: fair enough

arendjr (approved):

PR #154, opened 2024-01-16
Add new documentation site

laulauland:
# Description

This is the initial setup of the documentation site that includes the ported guides from the existing docs site as well as references generated directly from the OpenAPI spec. The CLI and templates references are for now directly copied from their respective folders - this will be updated to follow a similar pattern to be automatically generated from the source code later.

This direct access to our core components is the main motivation for moving the docs to this repo - it will be much easier to weave them into a well-documented experience for us internally and externally later.

### Architecture overview

The documentation site is an entirely static Astro site that uses Starlight framework for organizing and rendering documentation. I chose the Starlight framework as it provides very simple and good defaults for a documentation site: an interface for managing the sidebar and table of contents on each page, a search (built on Pagefind tool) that works out of the box, and a styling system that is accessible and looks good by default.

The entire site will be hosted in the `docs/` URL subdirectory as it will be hosted on the `fiberplane.com` main domain alongside our Framer-hosted site (as opposed to the `docs.*` subdomain). This is why you will find some of the files live in a somewhat awkward `docs/docs/` subdirectory - this is a work around to ensure that all markdown pages render on `fiberplane.com/docs` 

The site renders the documentation content from two sources:
1. Markdown files that are in the `src/content/docs/` subdirectory. These are pretty straightforward and they just render according to [Starlight's conventions](https://starlight.astro.build/guides/authoring-content/). As mentioned earlier the CLI and Templates reference are for now just these markdown files placed in their respective directories.
2. The OpenAPI spec that is rendered dynamically. This uses the underlying Astro's file-based routing engine to parse the `openapi_v1.yml` file and render the pages for each resource. For this I had to roll our own documentation components that I built on my own using the Starlight's design systems variables (with our fonts and colors mixed in) wherever possible. I could, however, make use of Starlight's template to make sure it sits flush with the rest of the documentation.

### How OpenAPI documentation is rendered

The documentation is organized using the `tags` properties in the OpenAPI spec. Each tag corresponds with what we would consider a "resource" in Fiberplane: notebooks, templates, workspaces, events etc. 

Each tag then has a number of operations associated that are rendered in a single large page that is ctrl/cmd-f'able along with necessary path, request body parameters, and return schemas. Every
path and operation makes use of the `description` property on the OpenAPI spec to render a rich markdown-supported documentation. The next phase in this project will entail a revision of the OpenAPI spec to ensure more of the relevant paths/parameters are better documented.

### Remaining TODOs and other notes

There is a number of ~~shortcuts~~ tradeoffs I made that will need to be addressed in subsequent PRs.

**Search**. Most of the search works out of the box thanks to Starlight. The parts that are not currently indexed are the API reference pages. This is the immediate next step after the first site is fully shipped.

**Blog site**. The blog site will eventually move to this website under the `blog/` subdirectory. 

**Types**. You'll notice there are maybe more type castings and `@ts-ignore` s then we'd otherwise be comfortable in our codebase 😅. I was a bit more liberal with them wanting to get the first version out the door. Also, all of the site is rendered at build time anyway for now.

**Design system**. The site mostly uses Starlight's defaults with Fiberplane system sprinkled in. Eventually we'll move to fully our own styling.

# Checklist

- ~~[ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~~


docs/biome.json
```diff
@@ -0,0 +1,25 @@
+{
```
flenter, line 1: Not a huge issue, but I noticed that there's a prettier (format) command in the package.json. Is biome not used here? The rest of the repo is using biome, right? Would it be possible to switch to biome?

docs/src/assets/images/1-creating-template.png
flenter, line 1: This is a big file, can't it be made a bit smaller? Perhaps even as a code block/snippet?

docs/src/assets/images/2-lambda-function.png
flenter, line 1: Another big image, I would try to limit the size of these images to be ideally something like 50KB or maybe 100KB? It's really easy to make them much smaller as JPEG if you don't need the transparency.

docs/src/lib/openApiSchema.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function groupOperationsByTag(schema: OpenAPIV3.Document): OpenAPIV3.Document {
+  const tagsMap = tagArrayToMap(schema.tags);
+  if (schema?.paths === undefined) {
+    console.error("No paths found in schema");
+    return schema;
+  }
+
+  for (const [path, obj] of Object.entries(schema?.paths) as [
```
stephlow: Maybe I'm missing something obvious, but are you sure this works? Or are you throwing intentionally? Since you're logging most of the time
```
Object.entries(undefined)
```
throws:
```
Uncaught TypeError: can't convert undefined to object
```

flenter (approved):

stephlow (approved):
Looking good!

Was wondering how hard it would be to set biome up with astro and its templating? Makes reviewing a bit easier if the code is formatted the same everywhere :)

docs/biome.json
laulauland: yeah i was looking for ways to avoid this but unfortunately Biome doesn't have Astro support yet (it doesn't parse HTML and CSS that make up the Astro components). Biome's own documentation site which is built on the same framework currently still uses Prettier for formatting.

laulauland:
@stephlow yeah unfortunately Biome doesn't support parsing Astro files as it doesn't yet understand HTML and CSS syntax. I was looking for a way to avoid adding Prettier but for now it seems like it's inevitable. Good news is that supporting HTML, CSS and their supersets like Astro seems to be on their [roadmap](https://biomejs.dev/blog/roadmap-2024/).

docs/src/assets/images/2-lambda-function.png
laulauland: These images get optimized at build time using [Astro's Image component](https://docs.astro.build/en/guides/images/#image--astroassets) so I generally don't worry about throwing larger images at it but good call on this and the other - I'm revisiting the guides content this week so I'll rework them.

laulauland: for example for these two images the build output is:

```
Size Name
56k dist/_astro/1-creating-template.CVhDXLuF_Z1pCE4.webp
108k dist/_astro/2-lambda-function.BkYztg7M_1mc2Jq.webp
```

docs/src/lib/openApiSchema.ts
laulauland: removed the redundant `?` as the there's already an undefined check

hatchan (approved):

PR #156, opened 2024-01-18
`@fiberplane/ui`: Simplify theme & provider

oscarvz:
# Description

- [x] No more styled-components theme values; only CSS variables
- [x] Greatly simplify `GlobalStyle` & only ship what's shared between Explorer & Studio
  - [x] Box element spacing values
  - [x] Media queries
- [x] Add variable fallback values
- [x] Documentation

# Checklist

~~- [ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~~


fiberplane-ui/README.md
```diff
@@ -7,17 +7,30 @@ Contains ReactJS UI elements for use in Fiberplane products.
...
+      type="submit"
       buttonStyle="secondary"
     >
       Click me!
     </Button>
   );
 };
 ```
+
+### Theming
```
arendjr: It seems this heading is missing content :)

arendjr (approved):
Looks like a nice improvement!

Does this mean we can also get rid of the `@fiberplane/ui/dist/theme` import soon?

fiberplane-ui/README.md
oscarvz: Whoops, that was drafted but will add it in the next iteration. Will remove 🧹

oscarvz:
@arendjr possibly, but we still want to serve an optional theme provider that we can use across our projects. The generated output from Figma doesn't contain any media query & sizing logic, but we follow the same approach across projects.
We definitely will get rid of `styled-components` as a dependency in this project however, so that'll clean up the `/theme` directory nicely. 😊

PR #158, opened 2024-01-22
[FP-3460] Add Front Matter manipulations to templates

gagbo:
# Description

- Add front matter support in templates (libsonnet)
- Fix name of data sources field in notebooks
- Fix bad interpolation of key name in front matter helper function
- Generate template docs

The extra functions in libsonnet allow to expand templates with pre-determined front matter

Fixes FP-3460

# Checklist

### DIFF
--- a/fiberplane-templates/docs/template_api.md
+++ b/fiberplane-templates/docs/template_api.md
@@ -41,7 +41,7 @@ fp.notebook.new('My Notebook')
         * [`.addFrontMatterCollection(name)`](#notebook.Notebook+addFrontMatterCollection) ⇒ [<code>Notebook</code>](#notebook.Notebook)
         * [`.addFrontMatterCollections(names)`](#notebook.Notebook+addFrontMatterCollections) ⇒ [<code>Notebook</code>](#notebook.Notebook)
         * [`.addFrontMatterSchema(frontMatterSchema)`](#notebook.Notebook+addFrontMatterSchema) ⇒ [<code>Notebook</code>](#notebook.Notebook)
-        * [`.addLabel(key, value)`](#notebook.Notebook+addLabel) ⇒ [<code>Notebook</code>](#notebook.Notebook)
+        * [`.addFrontMatterValue(key, value)`](#notebook.Notebook+addFrontMatterValue) ⇒ [<code>Notebook</code>](#notebook.Notebook)
         * [`.addFrontMatterValues(vals)`](#notebook.Notebook+addFrontMatterValues) ⇒ [<code>Notebook</code>](#notebook.Notebook)
     * [`.new(title)`](#notebook.new) ⇒ [<code>Notebook</code>](#notebook.Notebook)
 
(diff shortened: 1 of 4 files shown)


fiberplane-templates/fiberplane.libsonnet
```diff
@@ -400,6 +403,86 @@ local notebook = {
...
+     *
+     * @function notebook.Notebook#addFrontMatterCollections
+     * @param {Array.<string>} names - Names of the front matter collections in the workspace
+     * @returns {notebook.Notebook}
+     *
+     * @example notebook.addFrontMatterCollections(['post-mortem', 'opsgenie'])
+     */
+    addFrontMatterCollections(names):: std.foldl(
+      function(nb, name)
+        nb.addFrontMatterCollection(validate.string('name', name)),
```
arendjr: Nit: Looks like this gets validated twice :)

fiberplane-templates/fiberplane.libsonnet
```diff
@@ -400,6 +403,86 @@ local notebook = {
...
+      names,
+      self
+    ),
+
+    /**
+     * UNSTABLE: this function has no validation and the parameters might change
```
arendjr: I think this requires a newline. In the generated Markdown, it will run right into the next sentence.

arendjr (approved):

fiberplane-templates/docs/template_api.md
```diff
@@ -156,6 +166,85 @@ notebook.addLabels({
...
+
+**Kind**: instance method of [<code>Notebook</code>](#notebook.Notebook)  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| frontMatterSchema | <code>Array.&lt;object&gt;</code> | <p>Front Matter Schema as expected by the API</p> |
+
+<a name="notebook.Notebook+addLabel"></a>
+
+#### `notebook.addLabel(key, value)` ⇒ [<code>Notebook</code>](#notebook.Notebook)
```
hatchan: The header has not been updated.

fiberplane-templates/fiberplane.libsonnet
```diff
@@ -261,9 +261,12 @@ local notebook = {
   new(title):: {
     title: validate.string('title', title),
     timeRange: { minutes: -60 },
-    dataSources: {},
+    selectedDataSources: {},
```
hatchan, line 264: Is this supposed to change?

hatchan (commented):
I wonder if we can add some tests as well, to see if a specific template will be able to serialize to a `NewNotebook`. If the `selectedDataSources` fix in this PR was part of a issue somewhere else, then that would have been caught earlier (probably).

fiberplane-templates/fiberplane.libsonnet
gagbo: This is a small change that I spotted while working on the rest of the PR: the `dataSources` field is never used in libsonnet, and is not an existing field in the `NewNotebook` model. While debugging I enabled the `deny_unknown_fields` serde option as a lint and it gave me the error.

gagbo:
> I wonder if we can add some tests as well, to see if a specific template will be able to serialize to a `NewNotebook`. If the `selectedDataSources` fix in this PR was part of a issue somewhere else, then that would have been caught earlier (probably).

That would not have been caught by the tests, as the code was behaving normally, it’s really just a linting issue. It would have become an error if we enforced `serde(deny_unknown_fields)` on the `models::NewNotebook` struct

fiberplane-templates/fiberplane.libsonnet
```diff
@@ -400,6 +403,87 @@ local notebook = {
...
+     * notebook unless the front matter _schema_ of the notebook has an entry for the given key.
+     *
+     * @function notebook.Notebook#addFrontMatterValue
+     * @param {string} key - Key of the front matter entry
+     * @param {string | number} value - Front matter value
+     * @returns {notebook.Notebook}
+     *
+     * @example notebook.addFrontMatterValue(key='status', value='Created')
+     */
+    addFrontMatterValue(key, value):: self {
```
laulauland, line 463: is there a particular reason to have separate `addFrontMatterValue` and `addFrontMatterValues` interfaces? IMO this might trip up -  if I want to pass a single value, it's just a list with one item

fiberplane-templates/fiberplane.libsonnet
```diff
@@ -400,6 +403,87 @@ local notebook = {
...
+    /**
+     * Add a single front matter collection to the notebook.
+     *
+     * @function notebook.Notebook#addFrontMatterCollection
+     * @param {string} name - Name of the front matter collection in the workspace
+     * @returns {notebook.Notebook}
+     *
+     * @example notebook.addFrontMatterCollection(name='post-mortem')
+     */
+    addFrontMatterCollection(name):: self {
```
laulauland, line 416: same as the `addFrontMatterValue` - is there a particular reason to have separate interfaces for single and multiple values

fiberplane-templates/fiberplane.libsonnet
gagbo: I followed what was done for the labels functions. I don’t know if one version is more used than the others so I didn’t make the choice and implemented both as well

fiberplane-templates/fiberplane.libsonnet
gagbo: Oh true, thanks for catching that!

PR #165, opened 2024-02-06
Fix typescript test on CI

flenter:
# Description

Recently the ~@fiberplane/ui~ @fiberplane/hooks package was added as a dependency for @fiberplane/charts. However this resulted in the typescript checks to fail on CI because the typescript definitions could not be found. 

This PR adds a build step


oscarvz (approved):
Not UI lib, but hooks! Thanks for the fix

arendjr (approved):

PR #168, opened 2024-02-07
`@fiberplane/hooks`: Feature hooks

oscarvz:
# Description

Adds a helper function that takes an array of feature names, and returns feature fully typed features hooks.
This allows for using the helper function like so:

```tsx
// src/hooks/features.ts
import { getFeatureHooks } from "@fiberplane/hooks";

export const { useFeature, useFeatures, useFeaturesFromSearchParams } =
  getFeatureHooks(["feature-1", "feature-2"]);

```

Also includes a fix for Rollup in `fiberplane-charts` as the hooks library wasn't listed as an external dependency.

# Checklist

~~- [ ] The changes have been tested to be backwards compatible.~~
~~- [ ] New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~~

### DIFF
--- a/fiberplane-charts/rollup.config.ts
+++ b/fiberplane-charts/rollup.config.ts
@@ -13,10 +13,11 @@ const config = defineConfig([
       compact: true,
     },
     external: [
+      "@fiberplane/hooks",
       "framer-motion",
+      "react-window",
       "react",
       "react/jsx-runtime",
-      "react-window",
       "styled-components",
       "throttle-debounce",
     ],
--- a/fiberplane-hooks/src/useEffectOnce.ts
+++ b/fiberplane-hooks/src/useEffectOnce.ts
@@ -1,0 +1,9 @@
+import { EffectCallback, useEffect } from "react";
+
+/**
+ * Run a callback once when the component mounts.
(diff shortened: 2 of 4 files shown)


fiberplane-hooks/src/useEffectOnce.ts
```diff
@@ -0,0 +1,9 @@
+import { EffectCallback, useEffect } from "react";
+
+/**
+ * Run a callback once when the component mounts.
+ * @param effect Callback to run once
+ */
+export function useEffectOnce(effect: EffectCallback) {
+  useEffect(effect, []);
```
arendjr: I would suggest extracting the empty array to a variable in the scope above. That way we don't allocate a new one for each call.

arendjr (approved):

fiberplane-charts/rollup.config.ts
```diff
@@ -13,10 +13,11 @@ const config = defineConfig([
       compact: true,
     },
     external: [
+      "@fiberplane/hooks",
```
flenter, line 16: 😅

flenter (approved):

PR #170, opened 2024-02-12
Strengthen typing of front matter values in notebooks

gagbo:
# Description

Add stronger types to front matter values as stored in notebook models, so that
both API and consumers (Studio, CLI, scripts) can have better checks and understanding
of the front matter.

The main addition is from the example code usage in the doc, reproduced below:

```rust
// An existing schema to check values against
let schema = FrontMatterSchemaEntry::builder()
    .key("foo")
    .schema(FrontMatterNumberSchema::builder()
        .display_name("A number field that accepts single numbers")
        .build())
    .build();


// Another value (that has the wrong type)
let bad_value_from_api = json!("2022-10-08T13:29:00.78Z");
assert!(schema.validate_value(bad_value_from_api).is_err());

// A value can be inferred as well if it strictly matches the expected type
let inferred_value_from_api = json!(42);
assert!(schema.validate_value(inferred_value_from_api.clone()).is_ok());

// For example this will fail
let not_inferred_value_from_api = json!("42");
assert!(schema.validate_value(not_inferred_value_from_api).is_err());
```

# Checklist

- [x] remove the FrontMatterEnum... types to only use new value types


hatchan (approved):

mellowagain (approved):

arendjr (approved):

fiberplane-models/src/realtime.rs
```diff
@@ -559,6 +559,13 @@ pub enum RejectReason {
     #[serde(rename_all = "camelCase")]
     InconsistentFrontMatter { message: String },
 
+    /// The front matter updates in the operation does not apply cleanly to the notebook.
+    ///
+    /// The most common occurrence of this will be trying to insert/modify values that don’t
+    /// match the schema given in the operation.
+    #[serde(rename_all = "camelCase")]
+    InvalidFrontMatterUpdate { message: String },
```
hatchan: Can't we send over a `FrontMatterValidationError`? Just like we are doing with `InvalidLabel`

hatchan (approved):
One small thing

PR #175, opened 2024-02-19
PagerDuty webhooks integration

hatchan:
- [x] Add url to model
- [x] Add Sorting to list endpoint
- [x] Fix path for create endpoint


fiberplane-models/Cargo.toml
```diff
@@ -13,10 +13,12 @@ version = "1.0.0-beta.10"
 crate-type = ["lib"]
 
 [features]
+axum = ["dep:axum"]
 clap = ["dep:clap"]
 fp-bindgen = ["base64uuid/fp-bindgen", "dep:fp-bindgen"]
 
 [dependencies]
+axum = { version = "0.6.1", optional = true }
```
hatchan: This should implicitly add the `axum = ["dep:axum"]` feature, right? :thinking:

hatchan: Yes it does, and it works :100:

fiberplane-openapi-rust-gen/src/generator.rs
```diff
@@ -185,9 +186,12 @@ fn add_dependencies(dependencies: &mut DepsSet, args: &GeneratorArgs) -> Result<
         "fiberplane-models".to_string(),
         fp_dependency("fiberplane-models", args, Vec::new()),
     );
-
     dependencies.insert("bytes".to_string(), Dependency::Simple("1".to_string()));
 
+    dependencies.insert("thiserror".to_string(), Dependency::Simple("1".to_string()));
```
hatchan, line 191: This is needed for the `ClientError`

fiberplane-openapi-rust-gen/src/generator.rs
```diff
@@ -185,9 +186,12 @@ fn add_dependencies(dependencies: &mut DepsSet, args: &GeneratorArgs) -> Result<
...
+
+    dependencies.insert("url".to_string(), Dependency::Simple("2".to_string()));
```
hatchan, line 193: This is needed to expose the `UrlParse` error

fiberplane-openapi-rust-gen/src/routes.rs
```diff
@@ -55,11 +57,12 @@ pub(crate) fn generate_routes(
 
     writeln!(writer, "}}\n")?;
 
-    for (endpoint, item) in paths {
-        // this is so ugly omg 😭
+    writeln!(writer, "impl ApiClient {{")?;
```
hatchan, line 60: Functions are now part of the `ApiClient`, not separate functions that require a `ApiClient`.

fiberplane-openapi-rust-gen/src/routes.rs
```diff
@@ -113,30 +116,43 @@ pub(crate) fn generate_routes(
...
+    path: &str,
     method: &str,
     operation: &Operation,
     shared_parameters: &[RefOr<Parameter>],
     writer: &mut BufWriter<File>,
     components: &Components,
 ) -> Result<()> {
+    let use_new_style = operation
+        .extensions
+        .get("x-new-style")
```
hatchan, line 136: `x-new-style` should be used for any future endpoints. It will require a typed error and uses `PagedVec` for array endpoints.

fiberplane/Cargo.toml
```diff
@@ -14,6 +14,7 @@ crate-type = ["lib"]
 
 [features]
 api-client = ["fiberplane-api-client"]
+axum = ["fiberplane-models?/axum"]
```
hatchan: I couldn't come up with a better name, other than `axum`. This is to add the IntoResponse to the typed error objects (which includes the status code).

fiberplane-api-client/src/clients.rs
```diff
@@ -20,43 +20,28 @@ pub fn default_config(
...
-        let url = self.server.join(endpoint)?;
-
-        Ok(self.client.request(method, url))
-    }
-
-    pub fn builder(base_url: Url) -> ApiClientBuilder {
-        ApiClientBuilder::new(base_url)
-    }
-}
+#[deprecated(note = "Use `crate::ApiClient` instead")]
```
mellowagain: probably do something like this so it makes more sense to the end user:

```suggestion
#[deprecated(note = "Use `fiberplane_api_client::ApiClient` instead")]
```

fiberplane-openapi-rust-gen/src/client_config.rs
```diff
@@ -53,6 +53,24 @@ pub(crate) fn generate_client_configs(servers: &[Server], src_path: &Path) -> Re
...
+    let file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .append(false)
+        .open(src_path.join("api_client.rs"))
+        .context("Failed to open or create api_client.rs file")?;
+
+    let mut writer = BufWriter::new(file);
+
+    writeln!(writer, "{}", include_str!("../files/api_client.rs"))?;
```
mellowagain, line 67: this is really nice

fiberplane-openapi-rust-gen/src/client_config.rs
```diff
@@ -53,6 +53,24 @@ pub(crate) fn generate_client_configs(servers: &[Server], src_path: &Path) -> Re
...
+
+    writer
+        .flush()
+        .context("Failed to flush output for `builder.rs`")
```
mellowagain:
```suggestion
        .context("Failed to flush output for `api_client.rs`")
```

fiberplane-openapi-rust-gen/src/routes.rs
mellowagain: only in the openapi file i assume? or do we also need to include it on api responses?

mellowagain (approved):

fiberplane-openapi-rust-gen/src/routes.rs
hatchan: The `x-new-style` is only for the openapi schema indeed. The API endpoints and client will use the `PagedVec` (for array's) and will return the typed errors.

fiberplane-api-client/src/clients.rs
hatchan: The only "problem" is that this depends if you are using it through the `fiberplane` "meta" crate or directly using the `fiberplane_api_client`. I guess we can put both suggestions there :thinking:

fiberplane-models/src/pagerduty.rs
```diff
@@ -0,0 +1,357 @@
...
+            PagerDutyReceiverCreateError::Auth(AuthError::Unauthenticated) => {
+                StatusCode::UNAUTHORIZED
+            }
+            PagerDutyReceiverCreateError::Auth(AuthError::Unauthorized) => StatusCode::FORBIDDEN,
+            PagerDutyReceiverCreateError::DuplicateName => StatusCode::BAD_REQUEST,
+        };
+
+        (status_code, body).into_response()
+    }
+}
```
gagbo: I think that instead of implementing `IntoResponse` directly we could make a method like `into_status_and_body` method that doesn’t rely on axum. This way we don’t have to add `axum` as a dependency of `fiberplane-models`.

Not having `axum` in `fiberplane-models` is good to avoid issues with different versions of `axum`. For example if a crate using `fiberplane-models` also use `axum@0.7`, then this whole `IntoResonse` implemention is unusable, whereas a `err.into_status_and_body().into_response()` would work, albeit clunky.

I don’t have strong opinions on the matter though

fiberplane-models/src/pagerduty.rs
```diff
@@ -0,0 +1,357 @@
...
+impl PagerDutyReceiverListSortFields {
+    #[inline]
+    pub fn to_sql(&self) -> &'static str {
+        match self {
+            PagerDutyReceiverListSortFields::Name => "name",
+            PagerDutyReceiverListSortFields::CreatedAt => "created_at",
+            PagerDutyReceiverListSortFields::UpdatedAt => "updated_at",
+        }
+    }
+}
```
gagbo: Shouldn’t the `IntoStaticStr` from strum be enough? Or at least be reused in this function?

fiberplane-models/src/paging.rs
```diff
@@ -0,0 +1,225 @@
...
+impl<T> IntoResponse for PagedVec<T>
+where
+    T: Serialize,
+    Json<Vec<T>>: IntoResponse,
+{
+    fn into_response(self) -> Response {
+        let (results, headers) = self.unpack();
+        (headers, Json(results)).into_response()
+    }
+}
```
gagbo: Here I think the axum dependency makes more sense, but that also means we might need to have an `axum-06` and an `axum-07` feature

fiberplane-openapi-rust-gen/src/routes.rs
```diff
@@ -342,18 +367,59 @@ fn generate_route(
...
+    if use_new_style {
+        match resolve(
+            ResolveTarget::Response(&operation.responses.default.as_ref()),
+            components,
+        )? {
+            Some(ResolvedReference::Responses(response)) => {
+                let json_media = response
+                    .content
+                    .get("application/json")
+                    .expect("default error should have application/json");
```
gagbo: The error message should contain the operation ID so it’s easier to act on

fiberplane-openapi-rust-gen/src/routes.rs
```diff
@@ -342,18 +367,59 @@ fn generate_route(
...
+                let schema = json_media
+                    .schema
+                    .as_ref()
+                    .ok_or_else(|| anyhow!("need a schema"))?;
```
gagbo: The error message should contain the operation ID so it’s easier to act on

gagbo (approved):
The `axum` feature makes sense, but I’m afraid of potential compatibility issues downstream if they use a different version than `0.6`

fiberplane-models/src/pagerduty.rs
hatchan: Yeah fair, tbh I don't think this implementation detail should be exposed in `fiberplane/fiberplane` at all. The reason why this is a separate function is that in some situations the sort field might not map directly to a sql column.

I'll remove this and just use the strum impl for now.

PR #188, opened 2024-03-20
Add support for axum 0.7

hatchan:
- **Add support for axum 0.7**
- **temporary disable axum_06**


fiberplane-ci/Cargo.toml
```diff
@@ -15,7 +15,7 @@ anyhow = "1"
 async-trait = "0.1"
 clap = { version = "4", features = ["derive", "env"] }
 duct = "0.13"
-octocrab = { version = "0.23", default-features = false, features = ["rustls"] }
+octocrab = { version = "0.35", default-features = false, features = ["rustls"] }
```
hatchan, line 18: This was due to a issue with our internal workspace and this version. Since this is mostly used internally anyway I think we can just bump it.

fiberplane-models/src/pagerduty.rs
```diff
@@ -118,25 +112,46 @@ pub enum PagerDutyReceiverCreateError {
     Auth(AuthError),
 }
 
+impl PagerDutyReceiverCreateError {
+    #[cfg(any(feature = "axum_06", feature = "axum_07"))]
+    pub fn status_code(&self) -> http::StatusCode {
```
hatchan: Not sure if we want to make this part of the public api, thoughts?

fiberplane-models/src/paging.rs
```diff
@@ -104,17 +97,17 @@ impl<T> PagedVec<T> {
         self.map(|value| Into::<B>::into(value))
     }
 
-    #[cfg(feature = "axum_06")]
-    pub fn unpack(self) -> (Vec<T>, HeaderMap) {
+    #[cfg(any(feature = "axum_06", feature = "axum_07"))]
```
hatchan, line 106: Since this only requires `http` we can also add a new `http` feature and have the axum features also enable that, thoughts?

hatchan: Actually both axum_06 and axum_07 depend on different http versions :(

fiberplane-models/src/pagerduty.rs
hatchan: Giving that it now is a different type depending on the feature flags, I think it is best to keep it private. That is until we get rid of the conflicting types and until we get a good reason to expose this fn.

gagbo (approved):

mellowagain (approved):

PR #203, opened 2024-04-11
Allow configuration of default number of xticks and yticks

brettimus:
# Description

Sometimes we want to give a hint to the chart regarding how many ticks it should use on the axes. 

This PR allows you to set these as props on `MetricsChart`, while falling back to the hard coded values we're currently using.

Additionally, if you only specify 2 ticks, it assumes you just want to see the max and min. Unfortunately, this doesn't work too well for the Y-axis, and for the X-axis, your rightmost tick label will probably get cut off for timestamps. However, it's better than specifying two ticks and only having one at the origin and one in the middle.

Fixes FP-3595

# Checklist

### DIFF
--- a/fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
+++ b/fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
@@ -27,6 +27,8 @@ type Props = {
   scales: Scales;
   shouldAnimateYScale?: boolean;
   tickFormatters: TickFormatters;
+  numXTicks: number;
+  numYTicks: number;
 };
 
 export const GridWithAxes = memo(function GridWithAxes({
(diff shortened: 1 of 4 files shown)


fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
```diff
@@ -127,6 +131,11 @@ export function getTicks(
   numTicks: number,
   getMaxAllowedTick: (ticks: Array<number>, maxValue: number) => number,
 ): Array<number> {
+  // If we only want two ticks, just render the min and max
+  if (numTicks === 2) {
+    console.log("getTicks: numTicks === 2");
```
actualwitch: you forgot a console.log here

actualwitch (approved):

fiberplane-charts/src/CoreChart/GridWithAxes/GridWithAxes.tsx
brettimus: you're rightttt - removing

PR #213, opened 2024-04-24
Refactor cargo-check action to cargo-deny

hatchan:


deny.toml
```diff
@@ -0,0 +1,49 @@
...
+deny = []
+
+[sources]
+unknown-registry = "deny"
+unknown-git = "deny"
+allow-registry = ["https://github.com/rust-lang/crates.io-index"]
+allow-git = []
+
+[sources.allow-org]
+github = ["fiberplane", "autometrics-dev"]
```
gagbo, line 49: I think we can only allow `fiberplane` here, but to keep all files consistent it makes sense to have both, I now understand your point 👍

gagbo (approved):

PR #216, opened 2024-05-02
Implement / move UI components for PagerDuty integration

stephlow:
# Description

- Styling for form validation according to the design system
- Moves `Select` to `@fiberplane/ui`, updates it to the new design system and implements dark mode fixes
- Moves `TabList` to `@fiberplane/ui`, making `Tab`s regular elements instead of `<Link />`s and fixes dark mode issues
- Move `stopPropagation` and `cancelEvent` to `@fiberplane/ui` utils
- Move `AttachedPopup` to `@fiberplane/ui`
- Drop our use of `useClickAway` from `react-use` and implement a custom one in `@fiberplane/hooks` with a similar behaviour (dropped `events` param)
- Move `isOpen` to `@fiberplane/hooks`
- Introduce `useDocumentEventHandler` to set global event listeners on the document

Fixes FP-3491
Fixes FP-2490
Fixes FP-3686
Fixes AM-148
Fixes AM-147

# Checklist

- [ ] ~~The OpenAPI schema and generated client have been updated.~~~
- [ ] ~~New models module has been added to api generator xtask~~
- [ ] ~~New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~~

Monofiber: https://github.com/fiberplane/monofiber/pull/370


fiberplane-ui/package.json
```diff
@@ -35,8 +39,11 @@
     "typescript": "^5.4.2"
   },
   "peerDependencies": {
+    "@popperjs/core": "^2.11.8",
```
oscarvz, line 43: 🎈👀

oscarvz (approved):
🚀

stephlow:
Marked the PR as draft again as this branch seems to break `nx` on the CI on `monofiber`

fiberplane-hooks/src/useClickAway.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Hook that triggers a callback when a click, touch, or specified keyboard events occur outside of the specified element.
+ *
+ * @param {RefObject<HTMLElement | null>} ref - A ref object pointing to the element to monitor for outside interactions.
+ * @param {(event: MouseEvent | TouchEvent | KeyboardEvent) => void} onClickAway - Callback function that gets executed when an outside click, touch, or key event is detected.
+ */
+export const useClickAway = (
+  ref: RefObject<HTMLElement | null>,
+  onClickAway: (event: MouseEvent | TouchEvent | KeyboardEvent) => void,
+) => {
+  const savedCallback = useRef(onClickAway);
```
flenter: There's no need to store the `onClickAway` function in a ref when used in combination with `useHandler`.

fiberplane-hooks/src/useDocumentEventHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param {(event: GlobalEventHandlersEventMap[TEventType]) => void} handler
+ * The function that will be called when the event occurs. The event type corresponds to the provided `eventType`.
+ *
+ * @example
+ * // Single event usage
+ * useDocumentEventHandler('click', (event) => console.log('Clicked!', event));
+ *
+ * @example
+ * // Multiple events usage
+ * useDocumentEventHandler(['mousedown', 'mouseup'], (event) => console.log(`Event: ${event.type}`, event));
```
flenter: Wouldn't this result in the `useEffect` below being triggered each render? The array is being redefined each render.

fiberplane-ui/rollup.config.ts
```diff
@@ -5,14 +5,22 @@ import { defineRollupSwcOption, swc } from "rollup-plugin-swc3";
...
+    input: "src/index.ts",
     output: {
       file: "dist/index.js",
       format: "es",
       sourcemap: true,
       compact: true,
     },
-    external: ["react", "react/jsx-runtime", "styled-components"],
+    external: [
+      "@fiberplane/hooks",
```
flenter: Maybe it would be good to read the package json and loop over all peer dependencies and mark those as external? That way we don't need to maintain basically the same list twice.

flenter (commented):
Nice!  I did have some feedback though

fiberplane-ui/rollup.config.ts
stephlow: Great idea!

fiberplane-hooks/src/useClickAway.ts
stephlow: Thank you, removed this!

fiberplane-hooks/src/useDocumentEventHandler.ts
stephlow: You're right, didn't notice it at first because most code was ported from `react-use` without much of a thought. 

I decided to just remove the array for now, you'll just have to call the hook twice. There's probably a better way to support arrays and I'm all for it but trying to keep the scope of this PR down (since the scope creep has been real in this one ha), since we only have this use case a single time in our code base

flenter (approved):
LGTM

hatchan (approved):

PR #219, opened 2024-05-03
Fix clippies 2024-05-03

hatchan:
- **cargo clippy --fix**
- **Resolve more clippies**

### DIFF
--- a/fiberplane-openapi-rust-gen/src/routes.rs
+++ b/fiberplane-openapi-rust-gen/src/routes.rs
@@ -343,12 +343,12 @@ fn generate_route(
                 if response_type.is_none() {
                     response_type = Some(ResponseType::Json);
                 }
-            } else if response.content.get("text/plain").is_some() {
+            } else if response.content.contains_key("text/plain") {
                 response_type = Some(ResponseType::Text);
                 write!(writer, "String")?;
             } else {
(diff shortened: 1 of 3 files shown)


fiberplane-openapi-rust-gen/src/routes.rs
```diff
@@ -343,12 +343,12 @@ fn generate_route(
...
                     response_type = Some(ResponseType::Json);
                 }
-            } else if response.content.get("text/plain").is_some() {
+            } else if response.content.contains_key("text/plain") {
                 response_type = Some(ResponseType::Text);
                 write!(writer, "String")?;
             } else {
                 // octet-stream should be `bytes::Bytes` so don't warn about it when we reach this fallback
-                if response.content.get("application/octet-stream").is_none() {
+                if response.content.contains_key("application/octet-stream") {
```
gagbo: We want a `!` there, if the `get` is_none, that means the content doesn’t have the key

gagbo (changes requested):
It seems that one `get->is_none` conversion got swapped

fiberplane-openapi-rust-gen/src/routes.rs
hatchan: Good catch :100:

gagbo (approved):

PR #222, opened 2024-05-08
Fp 3664 upgrade biome

flenter:
# Description

Use latest version of biome


mellowagain (approved):
api side loosk good

fiberplane-hooks/src/useThemeSelect.ts
```diff
@@ -51,7 +51,7 @@ export function useThemeSelect() {
     return () => {
       clearTimeout(listenerId);
     };
-  }, [theme, storedTheme]);
+  }, [theme, storedTheme, previousTheme]);
```
oscarvz, line 52: With this change, the exclusion comment on line 31 & 32 should be removed. It's a tricky one, between builds this issue pops up from time to time.

oscarvz (approved):
Apart from a tiny thingy, lgtm!

PR #232, opened 2024-05-27
Add support for creating provider cell in markdown

flenter:
# Description

Adding something like:

```
  ``` promql
  # fiberplane-provider-query
```

Will now result in a provider cell with the query parameter set to `node_network_transmit_bytes_total[30s]`

Also:

The syntax that is specified in the markdown is now passed to the code cell

Fixes FP-3734 

- [ ]  ~~The OpenAPI schema and generated client have been updated~~
- [ ]  ~~New models module has been added to api generator xtask~~
- [ ]  ~~New types/fields are [well-documented and annotated](/CONTRIBUTING.md#adding-types-and-their-annotations).~~

### DIFF
--- a/fiberplane-markdown/src/from_markdown/mod.rs
+++ b/fiberplane-markdown/src/from_markdown/mod.rs
@@ -123,8 +123,7 @@ impl<'a> MarkdownConverter<'a> {
                             cell_text.push_str(&content);
                             continue;
                         }
-                    } else if self.current_code_block_syntax.is_some() {
-                        let syntax = self.current_code_block_syntax.take().unwrap();
+                    } else if let Some(syntax) = self.current_code_block_syntax {
 
                         // Check for magic string that indicates a query/provider cell
                         if content
--- a/fiberplane-markdown/src/from_markdown/mod.rs
+++ b/fiberplane-markdown/src/from_markdown/mod.rs
@@ -14,6 +14,7 @@ use tracing::warn;
 #[cfg(test)]
 mod tests;
 
+const PROVIDER_CELL_DIRECTIVE: &str = "# fiberplane-provider-query\n";
 /// Convert Markdown to a Fiberplane notebook
 ///
 /// This parses the first heading as the notebook title and
@@ -125,13 +126,11 @@ impl<'a> MarkdownConverter<'a> {
                         }
                     } else if let Some(ref syntax) = self.current_code_block_syntax {
                         // Check for magic string that indicates a query/provider cell
-                        if content
-                            .trim_start()
-                            .starts_with("# fiberplane-provider-query\n")
+                        if content.trim_start().starts_with(PROVIDER_CELL_DIRECTIVE)
                             && syntax.to_lowercase() == "promql"
                         {
                             let query = content
-                                .strip_prefix("# fiberplane-provider-query\n")
+                                .strip_prefix(PROVIDER_CELL_DIRECTIVE)
                                 .unwrap()
                                 .strip_suffix('\n')
                                 .unwrap();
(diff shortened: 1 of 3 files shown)


fiberplane-markdown/src/from_markdown/mod.rs
```diff
@@ -118,8 +123,53 @@ impl<'a> MarkdownConverter<'a> {
                             cell_text.push_str(&content);
                             continue;
                         }
+                    } else if self.current_code_block_syntax.is_some() {
+                        let syntax = self.current_code_block_syntax.take().unwrap();
```
gagbo:
```suggestion
                    } else if let Some(syntax) = self.current_code_block_syntax {
```

fiberplane-markdown/src/from_markdown/mod.rs
```diff
@@ -118,8 +123,53 @@ impl<'a> MarkdownConverter<'a> {
...
+
+                        // Check for magic string that indicates a query/provider cell
+                        if content
+                            .trim_start()
+                            .starts_with("# fiberplane-provider-query\n")
```
gagbo: If it’s a magic string it should probably be a `const &str` in the library

fiberplane-markdown/src/from_markdown/mod.rs
```diff
@@ -118,8 +123,53 @@ impl<'a> MarkdownConverter<'a> {
...
+                                    .syntax(syntax.as_str())
+                                    .build()
+                            } else {
+                                CodeCell::builder().content(content.to_string()).build()
+                            };
+
+                            // Create a Code cell
+                            self.new_cell(Cell::Code(code_cell));
+                        }
+                        self.current_code_block_syntax = None;
```
gagbo, line 168: Technically the `take` call will leave a None there, so the assignment here is not necessary. So either the `if let` just uses the `Option` like in the code suggestion, and you still need this assignment; or you do `if let Some(syntax) = self.current_code_block_syntax.take()` and you don’t need this extra assignment.

I tend to prefer the first solution, to avoid mutations in the branching conditions computation, but you do as you want

fiberplane-markdown/src/from_markdown/mod.rs
```diff
@@ -226,8 +276,18 @@ impl<'a> MarkdownConverter<'a> {
...
+                    CodeBlockKind::Fenced(language) => {
+                        self.current_code_block_syntax = Some(language.to_string());
+                    }
+                    CodeBlockKind::Indented => {
+                        self.current_code_block_syntax = Some("".to_string());
+                    }
+                }
+                // self.current_code_block_syntax = Some(kind.as_str());
+                self.current_cell = None;
+                // self.new_cell(Cell::Code(CodeCell::default()));
```
gagbo: No need for the comments there I think?

fiberplane-markdown/src/from_markdown/mod.rs
```diff
@@ -310,6 +370,7 @@ impl<'a> MarkdownConverter<'a> {
                 }
             }
             Tag::CodeBlock(_) => {
+                // self.in_code_block = false;
```
gagbo:
```suggestion
```

fiberplane-markdown/src/from_markdown/tests.rs
```diff
@@ -118,7 +118,7 @@ fn parsing_rich_text() {
 fn parsing_code_blocks() {
     let cells = markdown_to_cells(
         "
-```
+``` bash
 Some code
 ```
```
gagbo, line 123: If the test has to change, that probably means we need to make this version a major one.

flenter: The test didn't need to change. I modified an existing test case to test for an additional feature. I can also add a new test that specifically checks for this.

gagbo (approved):

actualwitch (approved):

